PR #54, opened 2015-06-02
[feat] Add per-tag scheme rules

### LINKED ISSUE #50: Allow for tag-specific schemes
Would like the ability to specify what schemes are supported on a per-tag basis. I could submit a pull request if you are interested.

For example:
```javascript
sanitizeHtml.defaults = {
  allowedTags: ['h3', 'h4', 'h5', 'h6', 'blockquote', 'p', 'a', 'ul', 'ol', 'nl', 'li', 'b', 'i', 'strong', 'em', 'strike', 'code', 'hr', 'br', 'div', 'table', 'thead', 'caption', 'tbody', 'tr', 'th', 'td', 'pre'],
  allowedAttributes: {
    a: ['href', 'name', 'target'],
    // We don't currently allow img itself by default, but this
    // would make sense if we did
    img: ['src']
  },
  // Lots of these won't come up by default because we don't allow them
  selfClosing: ['img', 'br', 'hr', 'area', 'base', 'basefont', 'input', 'link', 'meta'],
  // URL schemes we permit
  allowedSchemes: {
    a: ['http'],
    sanitizeDefault: ['http', 'https', 'ftp', 'mailto']
  }
};
```

cwill747:
This change allows scheme rules on a per-tag basis, in case you
want to limit the schemes that a certain tag can have, while not
enforcing the same restriction on another.

Closes #50

### DIFF
--- a/index.js
+++ b/index.js
@@ -131,7 +131,7 @@ function sanitizeHtml(html, options, _recursing) {
           if (!allowedAttributesMap || _.has(allowedAttributesMap[name], a) || (_.has(allowedAttributesGlobMap, name) &&
               allowedAttributesGlobMap[name].test(a))) {
             if ((a === 'href') || (a === 'src')) {
-              if (naughtyHref(value)) {
+              if (naughtyHref(name, value)) {
                 delete frame.attribs[a];
                 return;
               }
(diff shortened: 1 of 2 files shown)


boutell:
This looks much better. What's your use case for it?

boutell:
Sorry, you already said.

I think it's reasonable. I'm going to review a little and merge it.

cwill747:
No problem! We are really looking at the use case where we want users to be able to only have data: scheme's for  tags, but still allow http and https links for <a> tags. Just provides a bit more configuration if needed.

PR #65, opened 2015-07-21
using array.indexOf instead of array.includes

ArtskydJ:
Using `array.indexOf` instead of `array.includes`.

- Removed `array.includes` shim.
- Modified instances of `array.includes(whatever)` => `array.indexOf(whatever) !== -1`
- Added a few forgotten semicolons.
- Split a long line

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,12 +1,11 @@
 var htmlparser = require('htmlparser2');
 var extend = require('xtend');
 var quoteRegexp = require('regexp-quote');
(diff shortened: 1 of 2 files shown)


boutell:
Whoops, failing tests all to heck and back after this one. Backing it out.

ArtskydJ:
Whoops, forgot to check that before I made the PR. I'll see if I can track it down.

PR #68, opened 2015-08-08
Added the ability to transform every element using '*'

JamyDev:
The use case for this is if you want to validate a certain attribute on every element. For example, my usecase: running a CSS sanitize on every element's style attribute.

### DIFF
--- a/index.js
+++ b/index.js
@@ -73,11 +73,18 @@ function sanitizeHtml(html, options, _recursing) {
   });
 
   var transformTagsMap = {};
-  each(options.transformTags, function(transform, tag){
+  var transformTagsAll;
+  each(options.transformTags, function(transform, tag) {
+    var transFun;
     if (typeof transform === 'function') {
-      transformTagsMap[tag] = transform;
+      transFun = transform;
     } else if (typeof transform === "string") {
-      transformTagsMap[tag] = sanitizeHtml.simpleTransform(transform);
+      transFun = sanitizeHtml.simpleTransform(transform);
+    }
+    if (tag === '*') {
+      transformTagsAll = transFun;
+    } else {
+      transformTagsMap[tag] = transFun;
     }
   });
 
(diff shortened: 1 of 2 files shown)


JamyDev:
Also added a test, which passes.
```
$ npm test

> sanitize-html@1.7.2 test d:\WindowsData\Users\JamyDev\Programming\sanitize-html
> mocha test/test.js



  sanitizeHtml
    √ should be successfully initialized
    √ should pass through simple well-formed whitelisted markup
    √ should pass through all markup if allowedTags and allowedAttributes are set to false
    √ should respect text nodes at top level
    √ should reject markup not whitelisted without destroying its text
    √ should accept a custom list of allowed tags
    √ should reject attributes not whitelisted
    √ should accept a custom list of allowed attributes per element
    √ should clean up unclosed img tags and p tags
    √ should reject hrefs that are not relative, ftp, http, https or mailto
    √ should cope identically with capitalized attributes and tags and should tolerate capitalized schemes
    √ should drop the content of script elements
    √ should drop the content of style elements
    √ should preserve entities as such
    √ should dump comments
    √ should dump a sneaky encoded javascript url
    √ should dump an uppercase javascript url
    √ should dump a javascript URL with a comment in the middle (probably only respected by browsers in XML data islands, but just in case someone enables those)
    √ should not mess up a hashcode with a : in it
    √ should dump character codes 1-32 before testing scheme
    √ should dump character codes 1-32 even when escaped with padding rather than trailing ;
    √ should still like nice schemes
    √ should still like nice relative URLs
    √ should replace ol to ul
    √ should replace ol to ul and add class attribute with foo value
    √ should replace ol to ul, left attributes foo and bar untouched, remove baz attribute and add class attributte with foo value
    √ should replace ol to ul and replace all attributes to class attribute with foo value
    √ should replace ol to ul and add attribute class with foo value and attribute bar with bar value
    √ should skip an empty link
    √ Should expose a node's inner text and inner HTML to the filter
    √ Should collapse nested empty elements
    √ Exclusive filter should not affect elements which do not match the filter condition
    √ should disallow data URLs with default allowedSchemes
    √ should allow data URLs with custom allowedSchemes
    √ should allow specific classes when whitelisted with allowedClasses
    √ should allow defining schemes on a per-tag basis
    √ should not act weird when the class attribute is empty
    √ should not crash on bad markup
    √ should not allow a naked = sign followed by an unrelated attribute to result in one merged attribute with unescaped double quote marks
    √ should not filter if exclusive filter does not match after transforming tags
    √ should filter if exclusive filter does match after transforming tags
>>> √ should allow transform on all tags using '*'
    √ should not be faked out by double <
    √ should allow attributes to be specified as globs
    √ should quote regex chars in attributes specified as globs
    √ should not escape inner content from non-text tags (when allowed)
    √ should process text nodes with provided function


  47 passing (104ms)
```

PR #72, opened 2015-08-20
Update Readme to latest default options

emorikawa:
I noticed that the default options in the Readme were out of date. I updated them with the latest from the source code.


boutell:
Thanks.

PR #73, opened 2015-08-27
Feature: adding option to allow list of attributes to any tag.

zdravkoggeorgiev:
When I used sanitizer with whitelist option, I run into situation, when i wanted to allow a LOT of tags, and a LOT of attributes. But it was a pain to describe attributes for each of tags. So i made small changes, to allow some attributes to be allowed on any tag, declaring them like:

```javascript
allowedAttributes: {
  '*': [ 'href', 'align', 'alt', 'center', 'bgcolor' ]
}
```
It works, and it's handy for me. I am reconsidering to make some default whitelists too, because default whitelist is a way too restricted for HTML editors, for example.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,6 +83,15 @@ allowedAttributes: {
   a: [ 'href', 'data-*' ]
 }
 ```
+
+Also you can use the `*` as name for a tag, to allow listed attributes to be valid for any tag:
+
+```javascript
+allowedAttributes: {
+  '*': [ 'href', 'align', 'alt', 'center', 'bgcolor' ]
+}
+```
+
 ### htmlparser2 Options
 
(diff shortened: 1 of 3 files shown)


boutell:
Please add unit test coverage and verify all tests pass. Thanks!

zdravkoggeorgiev:
Thanks for the fast answer, and for sanitizer itself !
I will add unit test, but it will take few days, because I got a load at work.

zdravkoggeorgiev:
Added unit tests, they pass. Everything seems OK :)

Martii:
Seems well here in our sites default testing... thank you both. :)

boutell:
Thanks to zdravkoggeorgiev!

PR #84, opened 2016-01-05
class with WILDCARD

Rokt33r:


nelsonpecora:
I could really use this! (Plus it makes the API more consistent)

mheiber:
We could really use this, too! @Rokt33r can you take another look?

meijioro:
Yes please merge it.

index.js
```diff
@@ -77,6 +77,10 @@ function sanitizeHtml(html, options, _recursing) {
       allowedAttributesMap[tag].push('class');
     }
 
+    classes.map(function(clss) {
+      return quoteRegexp(clss).replace(/\\\*/g, '.*')
```
boutell: This doesn't look right. The return value of the map call is not used. Therefore classes is unmodified, unless I'm missing something. Has this been tested?

andrei-sheina:
Very necessary feature. #155

mheiber:
@nelsonpecora @meijioro @andrei-sheina @Rokt33r 
It looks like there may be a problem with the code in the PR. Is one of us taking a crack at fixing it? I can have a go, but don't want to duplicate effort.

index.js
nelsonpecora: Agreed. Removing this line, the tests still pass. I'm not sure what it's attempting to do here.

PR #85, opened 2016-01-12
Update htmlparser2

e-jigsaw:
`htmlparser2@3.8.x` has problem that crash with a specific character. Please update this module version.

Thanks

[Review on Reviewable](https://reviewable.io/reviews/punkave/sanitize-html/85)


boutell:
Thanks, I published an update to track htmlparser2's major version from now on as it is semver-compatible.

e-jigsaw:
@boutell Nice

PR #95, opened 2016-03-10
Pass stack (parent nodes) to textFilter()

biggeleben:
Would be great to have more context when processing text nodes, i.e. using textFilter(). Real-world use-case: Looking for plain-text URLs or mail addresses and turn them into clickable links. However, you don't want to do that if the parent node is already a link. Passing stack (actually in a reverse order is more convenient) would help a lot.

### DIFF
--- a/index.js
+++ b/index.js
@@ -205,7 +205,7 @@ function sanitizeHtml(html, options, _recursing) {
       } else {
         var escaped = escapeHtml(text);
         if (options.textFilter) {
-          result += options.textFilter(escaped);
+          result += options.textFilter(escaped, stack.slice().reverse());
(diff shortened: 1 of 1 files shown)


weixu365:
What's the status of this PR? I have the same requirement for a different reason.
I'm trying to clean up html, e.g. collapse whitespaces, ' \t ' should change to a single whitespace, but I want to skip the logic in pre/htmlarea tags, then I need to know what's the current and parent frame for the text

boutell:
A change like this should come with documentation changes explaining its use and giving an example.

abea:
Closing for inactivity.

PR #98, opened 2016-03-22
Allow for filtering of CSS Styles

cwill747:
Adds the ability to parse _inline_ css and decide whether or not to include it based off of the tag names. Uses the css library to parse out the inline css, filters it, then stringify's the CSS and re-inserts it, meaning that some whitespace or semicolons could be lost on the reparsing.

Definitely needs some code review.

### DIFF
--- a/README.md
+++ b/README.md
@@ -249,6 +249,29 @@ clean = sanitizeHtml(dirty, {
 });
 ```
 
+### Allowed CSS Styles
+
+If you wish to allow specific CSS _styles_ on a particular element, you can do that with the `allowedStyles` option.
+Any other CSS classes are discarded.
+
+This implies that the `style` attribute is allowed on that element
+
+```javascript
+clean = sanitizeHtml(dirty, {
+        allowedTags: ['p'],
+        allowedAttributes: {
+          'p': ["style"],
+        },
+        allowedStyles: {
+          '*': {
+            'color': '*',
+            'text-align': ['left','right','center','justify','initial','inherit'],
+            'font-size': '*'
+          }
+        }
+      });
+```
+
 ### Allowed URL schemes
 
 By default we allow the following URL schemes in cases where `href`, `src`, etc. are allowed:
--- a/index.js
+++ b/index.js
@@ -1,6 +1,8 @@
 var htmlparser = require('htmlparser2');
 var extend = require('xtend');
 var quoteRegexp = require('regexp-quote');
+var caja = require('html-css-sanitizer');
+var css = require('css');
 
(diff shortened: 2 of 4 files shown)


boutell:
Haven't had time to look at this but certainly very interested considering the XSS risks of both inline CSS and the style tag.

niftylettuce:
+1

pschwyter:
+1

mloschia:
Can someone merge this please?

cwill747:
Conflicts resolved, but the `css` npm project I depended on when I wrote this seems to be long-dead, and people are using postcss now. I'll refactor to use, since postcss is kept way more up to date

README.md
```diff
@@ -310,6 +310,29 @@ clean = sanitizeHtml(dirty, {
...
+
+```javascript
+clean = sanitizeHtml(dirty, {
+        allowedTags: ['p'],
+        allowedAttributes: {
+          'p': ["style"],
+        },
+        allowedStyles: {
+          '*': {
+            'color': '*',
```
pschwyter: can we allow users to specify a regexp? A wildcard defeats the point of having sanitization

pschwyter: probably *do not* want to  allow a wildcard

index.js
```diff
@@ -373,6 +388,41 @@ function sanitizeHtml(html, options, _recursing) {
...
+        });
+        originalAst.stylesheet.rules[0].declarations = allowedDeclarations;
+        return originalAst;
+      } else {
+        return false;
+      }
+    } else {
+      return originalAst;
+    }
+  }
```
mloschia: This method could be broken out into several smaller functions to make it more dry. I don't mind volunteering to do it.

README.md
mloschia: @pschwyter agreed, a wild card would introduce security concerns. That would enable anyone to enter malicious code

mloschia: RegExps would also enable whitelisting of measurement values. For example, if we wanted to whitelist pixels within font-size, all we'd need to do is append that to the regex.

index.js
mloschia: Also this is where we should do validation on the values of these css properties. It's be easy to do if Regexps are part of our whitelist

cwill747: I'll refactor on postcss and leave the function as-is, and then if you wanna chop it up, go for it!

mloschia: @cwill747 sounds good!

mloschia:
@cwill747 I did the package update and refactor. https://github.com/punkave/sanitize-html/pull/178

PR #108, opened 2016-05-14
Adds frame.children array

axdg:
This PR is in response to some undesired behaviour that I wasn't able to overcome with the current API.

Consider a situation where I want to allow anchors, but only when they either contain text nodes, or other nodes. Checking for `frame.text` in the `exclusiveFilter()` is generally sufficient - but will fail where all inner nodes are self closing tags (and hence don't contain text nodes).

So, for example:

``` js
sanitizeHtml('<a href="http://x.com"><img src="http://x.com/img.png" /></a>', {
    allowedTags: sanitizeHtml.defaults.allowedTags.concat([ 'img' ]),
    exclusiveFilter: function(frame) {
      return frame.tag === 'a' && !frame.text.trim();
    }
  }
); // => ''
```

This PR just adds an extra property (`children`) to the frame passed to `exclusiveFilter()`. `frame.children` is an array containing the tag names of any child nodes, it can be used to assert that a `frame` with empty `.text` contains self closing nodes.

It could also be used to further tune the behaviour of 'exclusiveFilter()', for example, to allow only anchors that contain a single image tag as per this [this test cases](https://github.com/punkave/sanitize-html/compare/master...axdg:master#diff-c1129c8b045390789fa8ff62f2c6b4a9R202).

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -186,6 +186,32 @@ describe('sanitizeHtml', function() {
             ''
         );
   });
+  it('Should not collapse elements containing only self closing tags', function() {
+    var tux = 'https://upload.wikimedia.org/wikipedia/commons/a/af/Tux.png?1463231285282';
+    var markup = '<a href="http://www.linux.org"><img src="'+ tux +'" /></a>'
+    assert.strictEqual(
+      sanitizeHtml(markup, {
+        allowedTags: sanitizeHtml.defaults.allowedTags.concat([ 'img' ]),
+        exclusiveFilter: function(frame) {
+          return (frame.tag === 'a') && !frame.text.trim() && !frame.children.length;
+        }
+      }),
+      markup
+    )
+  });
+  it('Should not collapse elements containing specified tags', function() {
+    var tux = 'https://upload.wikimedia.org/wikipedia/commons/a/af/Tux.png?1463231285282';
+    var markup = '<a href="http://www.linux.org"><img src="'+ tux +'" /></a>'
+    assert.strictEqual(
+      sanitizeHtml(markup, {
+        allowedTags: sanitizeHtml.defaults.allowedTags.concat([ 'img' ]),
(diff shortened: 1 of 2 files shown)


satyavh:
When is this going to be merged? This is THE missing feature!

satyavh:
@axdg This works but only for single nested children. So the following doesn't work and results in `children: []` You might want to check

<p><span></span</p>

axdg:
@satyavh, I'm not entirely sure what you mean... The PR creates an array of the children's tags names, so it does keep track of multiple children. If these children themselves have children, then they can be checked in their own filter callback.

satyavh:
What I mean is, this doesn't work:

```
<p>
  <span>
  </span>
  <span>
    <img ..../>
   </span>
</p>
```

when you log this out:
```
exclusiveFilter: (frame) ->
   console.log frame
```
then children:[] never contains the `img` tag.

I'm not sure what goes wrong, can't really put my finger on it.

axdg:
In the above case (where we are in the `<p>`'s callback), the children array will not contain the image tag, as it is not a child of the paragraph. The second spans callback can be used to check that the image tag is present though...

abea:
Sorry for the delay, everyone. I'm going to pick this back up. This would help resolve #313 as well.

test/test.js
```diff
@@ -186,6 +186,32 @@ describe('sanitizeHtml', function() {
             ''
         );
   });
+  it('Should not collapse elements containing only self closing tags', function() {
```
abea, line 189: This doesn't quite test for what it says it does. It doesn't matter if the tag is self closing. That can be fixed with a pretty small change in `index.js`, moving `frame.updateParentNodeChildren();` into the `options.Selfclosing` check below where it is now.

abea:
I'm closing this since the merge conflicts are too severe. The `index.js` file was moved to another directory. Please follow #313 for progress.

PR #130, opened 2016-09-08
Includes dist files for web applications

bennettmt:
Includes dist files for web applications that need to import from package managers


boutell:
Apologies for the absurd delay on this one. Actually though, we are distributing dist files already at this point. In fact we are overdoing it - we are using them as the "main" for those who do use a package manager, which is incorrect and causing some problems as noted elsewhere.

PR #138, opened 2017-01-18
Implemented allowedDomains option

surdu:
This will allow one to define an array with allowed domains in hrefs


surdu:
Started working on allowedDomains per tag name, please don't merge yet (if you intend to 😄 )

surdu:
Done 😄

boutell:
Thanks for this code!

I believe it is necessary to configure both www.youtube.com and youtube.com if we want to allow both, correct? I am not suggesting that we should try to figure that out automatically, but I do think that the option should be called allowedHostnames instead so people are not confused and don't expect it to detect that automatically.

Also, please add documentation (or if you prefer I can add it to the README).

surdu:
If you add youtube.com it will allow youtube.com and all it's subdomains, including `www`. I had all the "allow" spec variants that I removed thinking they would be redundant. Maybe I should add them back ... (https://github.com/punkave/sanitize-html/pull/138/commits/b3f0a3f46263e4e2db9c7b62c8659db8227eab52)

I could rename the option if you think it will be clearer for the user.

surdu:
OK, so I put back the deleted tests as I realize they can be useful both for making sure the functionality works as intended in the future and also for documentation purposes.

Considering that the option does allow both www.youtube.com and youtube.com if only youtube.com is specified do you still consider we should rename the option ?

surdu:
Is this PR still of interest ?

abea:
Closing for inactivity.

PR #156, opened 2017-08-11
Escape text from custom transformTags functions.

mikesamuel:
This makes custom tag transformations less error-prone.
Prior to this patch, tag transformations which turned an attribute
value into a text node could be vulnerable to code execution.

The operative change prevents any Frame's innerText from specifying
tag tokens.

----

Without this patch,

```js
sanitizeHtml(
    '<input value="&lt;script&gt;alert(1)&lt;/script&gt;">',
    {
      allowedTags: [ 'u' ],
      allowedAttributes: { '*': ['class'] },
      transformTags: {
        input: function (tagName, attribs) {
          return {
            tagName: 'u',
            attribs: { class: 'inlined-input' },
            text: attribs.value
          };
        }
      }
  });
```

returns

`"<u class="inlined-input"><script>alert(1)</script></u>"`

This makes custom tag transformers unnecessarily error-prone.


boutell:
I follow your logic, but this is a feature to some, and so needs to be addressed in a backwards compatible way, i.e. via an option. It might make sense to make it the default in 2.x.

mikesamuel:
Backwards compatibility is good but needs to be bounded.

The API is clear that `innerText` is supposed to be text, and this change makes the implementation consistent with the documented contract.

Another way to solve the underlying problem would be to reapply the parser, recursively, to `innerText`.  That would be inconsistent with the documentation though since you would be treating `innerText` as more like `innerHTML`.

If there's to be an option, IMO, it should default to treating `innerText` as text, with an manual override to treat it as `innerHTML`, or a different property with the name `innerHTML`.

Just be aware, that neither the current buggy interpretation, nor reinvoking the parser is going to solve problems where the parser has to invoke special parsing rules, e.g. you're reparsing and including the subtree at `(subtree)` in

```html
<a href="outer">(subtree)</a>
```

and the reparsed text is `<a href="inner"></a>` which would produce

```html
<a href="outer"><a href="inner"></a></a>
```

which is HTML that htmlparser2 would probably never produce for a single input since links don't nest without an intervening scoping element.

abea:
We'll keep this open for migration to the likely upcoming 2.x.

abea:
@boutell I've switched this to targeting the `v2` branch. There's some clean up to do first, but it might be easy enough since I believe we won't need `index.js` in a `src` directory anymore.

ggkitsas:
Hi, when do you plan to have a new release that includes this fix?

abea:
This is already released, @ggkitsas. You will need to specify `sanitize-html@^2.x` (or may even `2.0.0-beta`. It is not in npm as the "latest" release since it's still in beta for a bit longer.

Bronsequences:
Any idea when v2.0.0 will be out of beta? Our security scanning is flagging versions <2.0.0-beta as having a high severity security vulnerability `Affected versions of this package are vulnerable to Arbitrary Code Execution. Tag transformations which turn an attribute value into a text node using transformTags could be vulnerable to code execution.` which we would like to address. Thanks!

abea:
@Bronsequences It's in a [release candidate stage now](https://www.npmjs.com/package/sanitize-html/v/2.0.0-rc.2) (`2.0.0-rc.2`), if that matters. If you have identified security issues with that version, please open a new PR (or email security@apostrophecms.com) so we can resolve that and get it to the `2.0.0` version sooner. Or if the security flag is simply because it's been labeled a beta version, please confirm that as well.

If the current release candidate version looks good to you we might be pretty close to publishing that fully.

Also, great username.

abea:
Please see https://github.com/apostrophecms/sanitize-html/issues/376#issuecomment-689693382

Bronsequences:
Thanks @abea the security issue was flagged in the 1.x.x line with recommendations to upgrade to 2.0.0. We just wanted to wait until it was out of Beta before upgrading. Will use the rc version now that's available.

soljarjb13:
v

soljarjb13:
v

soljarjb13:
v

soljarjb13 (commented):
b

soljarjb13 (commented):
v

PR #157, opened 2017-08-23
allow classes with wildcards (update to #84)

### LINKED ISSUE #82: Support wildcards in allowedClasses configuration parameter
It would be nice to have possibility to wildcard classes like already done in "Wildcards for attributes" feature

```
allowedClasses: {
   '*': ['test-class', 'some-prefixed-*']
}
```

nelsonpecora:
This PR removes extraneous code from #84.

Closes #82

### DIFF
--- a/index.js
+++ b/index.js
@@ -303,7 +303,9 @@ function sanitizeHtml(html, options, _recursing) {
     classes = classes.split(/\s+/);
     return classes.filter(function(clss) {
       return allowed.some(function(_allowed) {
-        return clss.match(_allowed)
+        var allowedRegex = new RegExp('^' + _allowed.split('*').join('.*') + '$');
+
+        return clss.match(allowedRegex);
       })
     }).join(' ');
   }
--- a/index.js
+++ b/index.js
@@ -303,7 +303,7 @@ function sanitizeHtml(html, options, _recursing) {
     classes = classes.split(/\s+/);
     return classes.filter(function(clss) {
       return allowed.some(function(_allowed) {
-        var allowedRegex = new RegExp('^' + _allowed.split('*').join('.*') + '$');
+        var allowedRegex = new RegExp('^' + _allowed.split('*').join('.+') + '$');
 
         return clss.match(allowedRegex);
       })
--- a/index.js
+++ b/index.js
@@ -77,6 +77,10 @@ function sanitizeHtml(html, options, _recursing) {
       allowedAttributesMap[tag].push('class');
     }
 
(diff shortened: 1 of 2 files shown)


nelsonpecora:
@boutell hey, got a moment to look this over?

mheiber:
👀 this PR would help us

boutell:
Your code is treating the class name as a regular expression directly. The * works in your unit test because * happens to mean "zero or more repetitions of the preceding thing" which is the "-". But it will also match "nifty" without a "-" because that's still "zero or more repetitions."

Your original approach of correctly escaping the string as a regexp and then replacing an escaped "*" with ".*" was close, you just weren't using `map()` correctly. It doesn't modify all the elements of the array, it returns a new array with the results of the transformation. Assigning that to `classes` would probably get you to the finish line.

Here's a new test which currently fails if added to your code.

```javascript
it('Allowed class WILDCARD should match anything following that point, not just match preceding character in pattern zero or more times', function() {
    assert.equal(
      sanitizeHtml(
        '<p class="nifty simple dippy">whee</p>',
        {
          allowedTags: [ 'p' ],
          allowedClasses: {
            p: [ 'nifty-*' ]
          }
        }
      ),
      '<p class="">whee</p>'
    );
  });
```

nelsonpecora:
Thanks! I'll add that test and update my code in the next few days.

nelsonpecora:
@boutell sorry it took so long! I added your test and refactored the class matching to do proper wildcard regex.

nelsonpecora:
Hmm, it seems more logical for wildcards to match 1+, rather than 0+ (e.g. `foo-*` should match `foo-bar`, but not `foo-`), but it looks like attribute wildcards are matching 0+. What do you think?

mdavoodi:
I'd love to see this!

index.js
```diff
@@ -302,7 +302,11 @@ function sanitizeHtml(html, options, _recursing) {
     }
     classes = classes.split(/\s+/);
     return classes.filter(function(clss) {
-      return allowed.indexOf(clss) !== -1;
+      return allowed.some(function(_allowed) {
+        var allowedRegex = new RegExp('^' + _allowed.split('*').join('.+') + '$');
```
boutell, line 306: Apologies for sitting on this for a criminally long period of time!

There does still appear to be an issue, the wildcard is being set up properly but the actual text to be matched is not being escaped properly for use in a regular expression. Something like the regexp-quote module needs to be used to finish this out.

PR #166, opened 2017-10-02
Add enforceHtmlBoundary option to allow filtering out text outside html, and extra param to textFilter.

### LINKED ISSUE #149: Strip characters after </html>
Word 2010 sometimes adds strange chars after `</html>` e.g. the unicode replacement character (`\ufffd`)
I am manually removing them with `replace(/<\/html>[\s\S]*/, '</html>')` for now

### LINKED ISSUE #164: garbage text not cleaned up
I have a dirty html string that looks like this:

`"<html><body><p> some text </p></body></html> &%^junkunprintablecharshere"`

This string results from copy in MS outlook sometimes. I would expect sanitizeHtml to drop everything after closing `</html>` but it does not behave this way and I see the garbage chars after `</html>` returned as well. 

Is this by design and are there are ways to ensure clean html in returned i.e. all garbage after `</html>` is truncated when sanitizeHtml returns?

Many thanks.

andpor:
Add enforceHtmlBoundary option to allow filtering out text outside html tag boundary when present. This option is OFF by default so it would not impact anything. 

Resolves #164 
Resolves #149

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ node_modules
 dist
 # We do not commit CSS, only LESS
 public/css/*.css
+workspace.xml
--- a/README.md
+++ b/README.md
@@ -115,7 +115,8 @@ selfClosing: [ 'img', 'br', 'hr', 'area', 'base', 'basefont', 'input', 'link', '
 // URL schemes we permit
 allowedSchemes: [ 'http', 'https', 'ftp', 'mailto' ],
 allowedSchemesByTag: {},
-allowProtocolRelative: true
+allowProtocolRelative: true,
+enforceHtmlBoundary: false
 ```
 
 #### "What if I want to allow all tags or all attributes?"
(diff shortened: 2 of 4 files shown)


andpor:
Added documentation on the new option.

boutell:
Thanks for this. Two concerns:

1. You added an initializeState function, containing existing initialization code we formerly ran at all times, and made it conditional so that it only runs if the new flag is turned on. That doesn't appear to make sense. Why was this function added?

2. Please submit a clean PR without whitespace changes that make it more difficult to distinguish what is new in the patch.

Thanks!

andpor:
initializeState does not run conditionally. It is executed in the next statement right after it is defined, this is as it was. The reason I made it a function is that I did not want to duplicate the code in order to reset the state when condition to do so was detected. Basically the module starts off parsing and accumulating text until it sees opening html tag. When it see it and checks the property and it is true it basically resets the state as if the parsing has just been started essentially dropping everything it had done so far.

If this is not a good approach let me know.

I will do another PR without space changes...

andpor:
@boutell - this PR is now clean - all cosmetic changes were reverted. It is ready for your review.

andpor:
@boutell 

Tom - I added two more enhancements - one liners but really powerful. In transformTags currently you can return new text for the tag but you do not know what current text is , therefore opportunity for conditional cleanup for example does not exist. With extra text param you can have condition processing. Similarly for textFilter, current you just get text without any context what so ever so opportunity for any kind of conditional processing does not exist. If you pass the tag as second parameter, suddenly you have a world of opportunities to clean up you html string. I use it to remove extra spaces that sit between tags inside body which otherwise get returned in final result.

Hope you like this and will merge soon. All tests pass.

andpor:
I reverted the change related to passing the extra frame's text param to transformTags since it is always empty at that point.

andpor:
This is PR is clean, ready for review and merge...

Issue #57, opened 2015-06-30
escape instead of removing


iliakan:
I remember there was a way to escape instead of removing, but can't find in the docs.

Like: I have a `text <with> <tags>`, and I want it to become `text &lt;with&gt; &lt;tags&gt;`, all tags in-place, but escaped.

boutell:
I don't recall such a feature.

On Tue, Jun 30, 2015 at 7:52 AM, Ilya Kantor <notifications@github.com>
wrote:

> I remember there was a way to escape instead of removing, but can't find
> in the docs.
>
> Like: I have a text <with> <tags>, and I want it to become text
> &lt;with&gt; &lt;tags&gt;, all tags in-place, but escaped.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/punkave/sanitize-html/issues/57>.
>



-- 


*THOMAS BOUTELL, *DEV & OPS
P'UNK AVENUE | (215) 755-1330  |  punkave.com

iliakan:
So, there's no way to kraft it from existing stuff?

Looks like a fairly basic thing.

MauriceZ:
Will this be added any time soon?

boutell:
What is the use case where you want some tags to be respected, but others
to be escaped?


On Sun, Sep 6, 2015 at 12:23 PM, Maurice Zhang <notifications@github.com>
wrote:

> Will this be added any time soon?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/punkave/sanitize-html/issues/57#issuecomment-138097590>
> .
>



-- 


*THOMAS BOUTELL, *DEV & OPS
P'UNK AVENUE | (215) 755-1330  |  punkave.com

MauriceZ:
For example, if a user inputs a script tag, I would want it to appear rather than just deleting it to avoid confusion.

iliakan:
For example, I allow in comments `<a>` and `<i>` markup, but want other tags to appear "as is".

boutell:
It's a pull request I would consider. It doesn't make sense when sanitizing
HTML from a rich text editor etc., where people paste markup from word
processors that it would be better to just discard, but it could make sense
in a context where people are hand-writing markup I suppose.

On Sun, Sep 6, 2015 at 12:52 PM, Ilya Kantor <notifications@github.com>
wrote:

> For example, I allow in comments <a> and <i> markup, but want other tags
> to appear "as is".
>
> —
[...]

PR #169, opened 2017-10-15
Escape disallowed tags

yehonatanz:
Added the option to escape disallowed tags instead of stripping or discarding them. 
Escaping is a rather common use-case (a user naively sending `<span>hello</span>` would expect it to render like that, and not stripped down to `hello`), which is also referenced in #57 .

### DIFF
--- a/README.md
+++ b/README.md
@@ -488,6 +488,16 @@ Note that if you use this option you are responsible for stating the entire list
 
 The content still gets escaped properly, with the exception of the `script` and `style` tags. *Allowing either `script` or `style` leaves you open to XSS attacks. Don't do that* unless you have good reason to trust their origin.
 
+### Escaping the content of a disallowed tag
+
+Instead of discarding, or keeping text only, you may enable escaping of the entire content:
+
+```javascript
+escapeDisallowedTags: true
+```
+
+This will transform `<disallowed>content</disallowed>` to `&lt;disallowed&gt;content&lt;/disallowed&gt;`
+
 ## About P'unk Avenue and Apostrophe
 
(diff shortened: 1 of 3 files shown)


yehonatanz:
@boutell 
Hey, I have rebased my PR so it's no longer conflicting.
Would appreciate some feedback.

boutell:
Hmm, I don't get why you would want to escape non-whitelisted child elements. I would not expect that, if I was a semi-naive HTML user "hoping for the best."

martinmckenna:
Coincidentally just posted an issue about this exact thing.

I would love to render blacklisted tags as text. In fact, that's exactly what github does.

<script>alert('hello world')</script>

boutell:
I'm persuaded there's value here in this as an optional behavior activated by a flag, but it looks like the PR has some conflicts that need to be resolved.

yehonatanz:
Hey, rebased my changes again several days ago, but for some reason the circleci check is still waiting.
Any ideas?

martinmckenna:
Went ahead and created a PR to DefinitelyTyped to handle the TypeScript-friendly stuff.

martinmckenna:
@boutell any movement on this?

boutell:
I had to revert this. Your tests failed for me with "npm test". Indeed it doesn't seem to work.

Maybe you had trouble running "npm test" because of the issues that crept in with the frontend build. I just fixed that in master, so merge latest master into your branch and try again? Thanks.

yehonatanz:
As a matter of fact, I did have troubles running "npm test", but I ran the tests manually and everything seemed fine.
I'll have time to look into it and fix it in November.

yehonatanz:
Seems like `npm test` is broken on master in a way unrelated to this PR.
Please review #311 , if and when it is merged, I'll rebase this PR.

PR #172, opened 2017-10-30
Disallow URLs starting with /\ in case allowProtocolRelative is false

### LINKED ISSUE #171: External URLs can be used even when allowProtocolRelative is false
It's possible to smuggle trough 'protocol-relative' URLs even when `allowProtocolRelative` is disabled. 

Proof of concept:

```javascript
options = {
  allowedTags: [ 'a' ],
  allowedAttributes: {
    a: [ 'href'],
  },
  allowProtocolRelative: false
};

console.log(sanitizeHtml("<a href=\"//test.com\">Click me!</a>", options)); // <a>Click me!</a>
console.log(sanitizeHtml("<a href=\"/\\test.com\">Click me!</a>", options)); // <a href="/\test.com">Click me!</a>
```

Note that `/\test.com` is considered 'protocol-relative' by latest Chrome and FF. It should probably be enough to extend the regex check to match `/\` also.

bayotop:
Fixes #171

### DIFF
--- a/index.js
+++ b/index.js
@@ -296,8 +296,8 @@ function sanitizeHtml(html, options, _recursing) {
     // Case insensitive so we don't get faked out by JAVASCRIPT #1
     var matches = href.match(/^([a-zA-Z]+)\:/);
     if (!matches) {
-      // Protocol-relative URL: "//some.evil.com/nasty"
-      if (href.match(/^\/\//)) {
+      // Protocol-relative URL starting with any combination of '/' and '\'
+      if (href.match(/^[\/\\]{2}/)) {
         return !options.allowProtocolRelative;
       }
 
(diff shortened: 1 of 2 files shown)


boutell:
Good catch, however it looks like \\ works too, not just \/. Any combination of / and \ for the first two characters will do it. So your regex requires a little bit more tweaking. Got a moment to do that?

bayotop:
You are right. I amended the change to reflect that behavior (interestingly FF treats even `\\example.com` as protocol-relative).

boutell:
I think it's there to be tolerant of confused Windows developers. MSDOS is ever with us.

boutell:
Thanks for the fix!

PR #178, opened 2017-11-13
CSS Sanitization

mloschia:
This is an updated pull request from @cwill747 [original](https://github.com/punkave/sanitize-html/pull/98)

The `allowedStyles` option now requires a list of regular expressions for whitelisting. This makes whitelisting attributes such as font-size: 30px much easier because all unit types don't have to be accounted for within our logic.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ HTML comments are not preserved.
 
 ## How to use
 
-### Browser 
+### Browser
 
 *Think first: why do you want to use it in the browser?* Remember, *servers must never trust browsers.* You can't sanitize HTML for saving on the server anywhere else but on the server.
 
@@ -312,7 +312,7 @@ clean = sanitizeHtml(dirty, {
 
 ### Allowed CSS Styles
 
-If you wish to allow specific CSS _styles_ on a particular element, you can do that with the `allowedStyles` option.
+If you wish to allow specific CSS _styles_ on a particular element, you can do that with the `allowedStyles` option. Simply declare your desired attributes as regular expression options within an array for the given attribute. Specific elements will inherit whitelisted attributes from the global (\*) attribute.
 Any other CSS classes are discarded.
 
 This implies that the `style` attribute is allowed on that element
@@ -325,9 +325,14 @@ clean = sanitizeHtml(dirty, {
         },
         allowedStyles: {
           '*': {
-            'color': '*',
-            'text-align': ['left','right','center','justify','initial','inherit'],
-            'font-size': '*'
+            // Match HEX and RGB
+            'color': [/\#(0x)?[0-9a-f]+/i, /rgb\(\s*(\d{1,3})\s*,\s*(\d{1,3})\s*,\s*(\d{1,3})\s*\)$/],
+            'text-align': [/left/, /right/, /center/],
+            // Match any number with px, em, or %
+            'font-size': [/\d+[px|em|\%]/]
+          },
+          'p': {
+            'font-size': [/\d+rem/]
           }
         }
       });
--- a/README.md
+++ b/README.md
@@ -329,7 +329,7 @@ clean = sanitizeHtml(dirty, {
             'color': [/\#(0x)?[0-9a-f]+/i, /rgb\(\s*(\d{1,3})\s*,\s*(\d{1,3})\s*,\s*(\d{1,3})\s*\)$/],
             'text-align': [/left/, /right/, /center/],
             // Match any number with px, em, or %
-            'font-size': [/\d+[px|em|\%]/]
+            'font-size': [/^\d+$[px|em|\%]/]
           },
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -310,6 +310,34 @@ clean = sanitizeHtml(dirty, {
...
+clean = sanitizeHtml(dirty, {
+        allowedTags: ['p'],
+        allowedAttributes: {
+          'p': ["style"],
+        },
+        allowedStyles: {
+          '*': {
+            // Match HEX and RGB
+            'color': [/\#(0x)?[0-9a-f]+/i, /rgb\(\s*(\d{1,3})\s*,\s*(\d{1,3})\s*,\s*(\d{1,3})\s*\)$/],
+            'text-align': [/left/, /right/, /center/],
```
pschwyter: since we can represent any of these rules with regexps, we don't need to define the rules as an array, just regexp itself is fine

index.js
```diff
@@ -373,6 +393,88 @@ function sanitizeHtml(html, options, _recursing) {
...
+            return objValue.concat(srcValue);
+          }
+        }
+      );
+    } else {
+      selectedRule = allowedStyles[astRules.selector] || allowedStyles['*'];
+    }
+
+    /**
+     * Filters the existing attributes for the given property. Discards any attributes
```
pschwyter: Did you mean to include this for the `astrules.nodes.reduce` callback?

Function docblocks shouldn't be inline like this. It probably makes sense to move the callback out of this block and make it a named function.

boutell:
I'm not crazy about the idea of elaborate regexps that still (I think?) do not accept all of the valid syntax possibilities. Real CSS parsing would seem to be the right way to do this, if it's going to be done.

index.js
mloschia: Yeah I did. It's complex enough to warrant being moved out.

README.md
```diff
@@ -310,6 +310,34 @@ clean = sanitizeHtml(dirty, {
...
+            // Match any number with px, em, or %
+            'font-size': [/\d+[px|em|\%]/]
```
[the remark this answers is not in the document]
mloschia: I'll make it more extensive

README.md
mloschia: I'm fine either way with how it is, though keeping these as lists could yield a better developer experience

mloschia:
@boutell The regexps are declared by the developer in order to account for any syntax possibilities. The most up to date CSS parsers I found take the same approach to parsing the css as we did in this PR (using an abstract syntax tree), though the implementation in this repository is specifically used for whitelisting purposes. I think a css whitelisting library with flexibility is a good long-term solution however there has been a reoccurring need for this whitelisting feature to be part of this library specifically.

README.md
mloschia: I disagree. There are global style whitelistings which should be inherited by individual element whitelistings. Take a look at this example:

```javascript
        allowedStyles: {
          '*': {
            "color": [/yellow/],
            "text-align": [/center/]
          },
          'span': {
            "color": [/green/],
            "font-family": [/helvetica/]
          }
        }
```

This should enable all the properties on this span to pass:

```<span style="color:yellow;text-align:center;font-family:helvetica;"></span>```.

By only using regular expressions in these values we would have to either concatenate regular expressions or put the global and individual regular expressions in an array and check them individually.

Comment (pschwyter) [reply README.md]: ahh I see, that's a cool feature

**Comment by mloschia** [comment]
@boutell Thoughts? Can we merge this?

File: README.md [hunk 4]
```diff
@@ -310,6 +310,34 @@ clean = sanitizeHtml(dirty, {
 });
 ```
 
+### Allowed CSS Styles
+
+If you wish to allow specific CSS _styles_ on a particular element, you can do that with the `allowedStyles` option. Simply declare your desired attributes as regular expression options within an array for the given attribute. Specific elements will inherit whitelisted attributes from the global (\*) attribute.
+Any other CSS classes are discarded.
+
+This implies that the `style` attribute is allowed on that element
```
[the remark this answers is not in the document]
boutell: I'll clarify this after merge.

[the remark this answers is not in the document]
mloschia: The developer will have to enable the style attribute first, thanks for clarifying!

mloschia:
Thanks @boutell ! Yes I agree that there should be an exhaustive library for this. I also like the idea of being able to set `true` to enable all inputs.

Do you have a perspective date the next release will be?

boutell:
The feature has been released.

I would envision `true` enabling a standard regexp for that attribute. 

One concern I still have here is that we're not validating URLs that appear through the same mechanisms we use for all other URLs, correct? I did document that, but it's not ideal.

mloschia:
Great 👍

Correct, with the current implementation any URLs would have to be white listed via regex.

PR #182, opened 2017-11-22
Remove ES6 syntax

pschwyter:
ES6 syntax was introduced in PR #174, with the introduction of a single `const`. Until then, this package was fully ES5 compatible requiring no transpiling step for use in older (<IE11) browsers.

Since this use of `const` seems to be the exception, I'm hoping we can revert back simply using `var`.

In the future, if we want to introduce ES6 syntax then maybe we can take one of two approaches:

1. a major version change (if we want to drop support for older browsers)
2. include a transpile step in the browser build (if we want to continue supporting older browsers)

### DIFF
--- a/index.js
+++ b/index.js
@@ -42,7 +42,7 @@ module.exports = sanitizeHtml;
 //   * unexpected-character-in-attribute-name
 // We exclude the empty string because it's impossible to get to the after
 // attribute name state with an empty attribute name buffer.
-const VALID_HTML_ATTRIBUTE_NAME = /^[^\0\t\n\f\r /<=>]+$/;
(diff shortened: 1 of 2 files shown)


boutell:
I do think it's odd to go to ES6 just for one tiny change, except: what browser are you actually concerned about? Microsoft has dropped all security updates for IE less than 11. (I think there might be one obscure version of Windows Server - not desktop Windows - that still only goes up to 10 and hasn't been completely dropped yet, but I'm having trouble finding a reference.)

https://www.microsoft.com/en-us/WindowsForBusiness/End-of-IE-support

pschwyter:
Yeah, that's a fair point.

I am using this in a library where I need to support <IE11. However, if you don't intend to support <IE11 (and that makes sense when Microsoft themselves no longer even support it), then I would say it's up to the package user (ie. me) to transpile the package.

PR #183, opened 2017-11-23
replace lodash.isarray with Array.isArray

OmgImAlexis:
Fixes the deprecation issue.

```
sanitize-html > lodash.isarray@4.0.0: This package is deprecated. Use Array.isArray.
```

PR #184, opened 2017-11-27
remove lodash.isarray

graingert:


graingert:
oh woops there's already a PR: https://github.com/punkave/sanitize-html/pull/183/files

PR #190, opened 2017-12-08
Feature/babelify index.js

ayushya:
This PR babelify index.js

closes https://github.com/punkave/sanitize-html/issues/189


boutell:
Upon reflection, there's no need for me to nitpick about using babel in node. It broadens support for old versions of node at no real cost. Thanks for the work.

PR #191, opened 2017-12-13
note that using santize-html on content from servers is fine.

graingert:


graingert:
eg, imhop it's best to keep user data as it's entered for as long as possible, only sanitizing it at the last possible moment before it's inserted in the page.

if you santize content before it's inserted and blacklist `<em>` you can't then change the rules without un-santizing your database content and re-sanitizing. If you santize only when the content is inserted in the dom you can change your mind on supported tags.

boutell:
I don't buy this argument. You can store both the original and the sanitized version on the server if you want, and resanitize as a batch job at any time. Hundreds of kilobytes of extra javascript to do it in the browser on every insert into the page is not the sensible solution, although it is not irresponsible in the way that trusting the browser at submit time would be.

PR #196, opened 2018-01-08
add chalk to dependencies

webdif:
Following this discussion: https://github.com/punkave/sanitize-html/issues/195#issuecomment-355746502


boutell:
Wanted to confirm that this is tested - it generates a bundle that works for you?

webdif:
Yes, it works! 👌

boutell:
This was published to npm.

PR #202, opened 2018-01-30
Remove unused chalk dependency

vladgolubev:
1. It's unused
2. It breakes IE11

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "sanitize-html",
-  "version": "1.16.3",
+  "version": "1.17.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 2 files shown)


boutell:
Without this explicit dependency, we get reports that it breaks the build, because it is required by one of the CSS parser libraries... this is all browser stuff which is a very marginal use case for any kind of sanitization, but I'm about done trying to sort it out.

vladgolubev:
@boutell  is it possible to add it to devDependencies then? If it's used only for build?

boutell:
What I mean is that somebody else indicated we needed it or he couldn't use it in the browser.

https://github.com/punkave/sanitize-html/issues/195

We have been using babel as part of the process to build the frontend bundle for a while but apparently that doesn't currently babelize all of the dependencies? Perhaps if someone took that on.

¯\_(ツ)_/¯

vladgolubev:
LOL 😅 

image

I rushed w/o research, so closing this

PR #214, opened 2018-03-22
Fix regex in font-size filtering example in README.md

jasco:
Remove stray $ from the example filter for `font-size`.  Use grouping rather than charset for alternatives.

Note might opt to use capture rather than non-capture grouping to make consistent with other examples.


boutell:
Thanks.

Issue #215, opened 2018-04-04
The iframe src code sample from the Iframe Filters section of https://www.npmjs.com/package/sanitize-html is not working


iambrosie:
Hi,

I am trying to run the following code sample:
`clean = sanitizeHtml('<p><iframe src="https://www.youtube.com/embed/nykIhs12345"></iframe><p>', {`
`  allowedTags: [ 'p', 'em', 'strong', 'iframe' ],`
`  allowedClasses: {`
`    'p': [ 'fancy', 'simple' ],`
`    'iframe': ['src']`
`  },`
`  allowedIframeHostnames: ['www.youtube.com', 'player.vimeo.com']`
`});`

which is provided as a sample for using Iframe Filters here: https://www.npmjs.com/package/sanitize-html.
The output I get after running the aforementioned code sample is:

`me@my:~/path/to/folder$ node index.js`
`<p><iframe></iframe></p><p></p>`

The expected output is:
`<p><iframe src="https://www.youtube.com/embed/nykIhs12345"></iframe><p>`

The sanitize-html version used is 1.18.2, together with node version 1.8.2

Kind regards,
Ionut Ambrosie

boutell:
@ryan-veris would you please take a look at this since it's your contribution? Thanks.

ryan-verys:
Hi @iambrosie this may be a misprint in the docs, `src` is an attribute not a class. Instead of

`allowedClasses: {
'iframe': ['src']
},`

can you try 

`allowedAttributes: {
'iframe': ['src']
},`

and report your results?

ryan-verys:
Ok I'll update the docs. Sorry for the confusion!

iambrosie:
No worries. Glad it works now.

PR #216, opened 2018-04-05
Updated iframe whitelist misprint

ryan-verys:
This fixes misleading misprint in the README that shows that the iframe `src` should be added to `allowedAttributes`. 

In re: #215


boutell:
Thanks!

PR #245, opened 2018-08-02
Restore package-lock

### LINKED ISSUE #241: Wrong build 1.18.3? Uncaught TypeError: (0 , _sanitizeHtml2.default) is not a function
Hey guys, just received this when new version has been released.

```
    sanitizeHtml.js?70b5f51:7 Uncaught TypeError: (0 , _sanitizeHtml2.default) is not a function
    at sanitize (sanitizeHtml.js?70b5f51:7)
    at createSanitizeMarkup (sanitizeHtml.js?70b5f51:12)
    at TemplatePainterAboutView (View.js?f4af559:50)
    at mountIndeterminateComponent (react-dom.development.js?02e2fdd:8574)
    at beginWork (react-dom.development.js?02e2fdd:8978)
    at performUnitOfWork (react-dom.development.js?02e2fdd:11814)
    at workLoop (react-dom.development.js?02e2fdd:11843)
    at HTMLUnknownElement.callCallback (react-dom.development.js?02e2fdd:100)
    at Object.invokeGuardedCallbackDev (react-dom.development.js?02e2fdd:138)
    at invokeGuardedCallback (react-dom.development.js?02e2fdd:187)
```

Update: We use `import sanitizeHtml from 'sanitize-html'`
Update2: Affects FE build
Update3: in package.json until it's fixed
```
"sanitize-html": "^1.18.2"
// =>
"sanitize-html": "1.18.2"
```

### LINKED ISSUE #242: Version 1.18.3 does not have defaults defined
When working with version 1.18.2 I was able to access `sanitizeHtml.defaults` but with version 1.18.3 I'm getting undefined. If I rollback to the former version, there are no problems with the following line:
`allowedTags: sanitizeHtml.defaults.allowedTags.concat(['h1', 'h2'])`

felixfbecker:
Fixes #242 
Fixes #241

and prevents breakages like it in the future.

> This was caused by gitignoring and removing package-lock.json in https://github.com/punkave/sanitize-html/pull/240/files#diff-32607347f8126e6534ebc7ebaec4853d. What is the reason this was done? It seems that without the lockfile, some devDependency was silently updated, which changed the build output. This is exactly what lockfiles are there for to prevent - making the build reproducible and deterministic, so breakages can't suddenly happen without any changes to master. Please keep a lockfile checked in.

#244  should still be merged.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
-package-lock.json
 npm-debug.log
 *.DS_Store
 node_modules
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,4334 @@
+{
+  "name": "sanitize-html",
+  "version": "1.18.0",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
+    "JSONStream": {
+      "version": "1.3.1",
+      "resolved": "https://registry.npmjs.org/JSONStream/-/JSONStream-1.3.1.tgz",
+      "integrity": "sha1-cH92HgHa6eFvG8+TcDt4xwlmV5o=",
+      "dev": true,
+      "requires": {
+        "jsonparse": "1.3.1",
+        "through": "2.3.8"
+      }
+    },
+    "acorn": {
+      "version": "4.0.13",
+      "resolved": "https://registry.npmjs.org/acorn/-/acorn-4.0.13.tgz",
+      "integrity": "sha1-EFSVrlNh1pe9GVyCUZLhrX8lN4c=",
+      "dev": true
+    },
+    "align-text": {
+      "version": "0.1.4",
+      "resolved": "https://registry.npmjs.org/align-text/-/align-text-0.1.4.tgz",
+      "integrity": "sha1-DNkKVhCT810KmSVsIrcGlDP60Rc=",
+      "dev": true,
+      "requires": {
+        "kind-of": "3.2.2",
+        "longest": "1.0.1",
+        "repeat-string": "1.6.1"
+      }
+    },
+    "ansi-regex": {
+      "version": "2.1.1",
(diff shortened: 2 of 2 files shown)


boutell:
I didn't find that this was necessary. With the "browser" keyword removed, I get a good build even if I remove package-lock.json. And it never makes sense to suggest package-lock.json is going to be honored in a library because it won't be (project level npm install will ignore any package-lock other than its own). So I think this can be closed, but let me know if 1.18.4 doesn't do the trick for you.

felixfbecker:
It won't be honored for dependencies, which is why dependencies should be updated often (best automatically), but it will be for devDependencies. As a user, it makes me nervous to use this library if between releases a bug in e.g. Babel or Browserify could cause a faulty build.

If the bug was introduced by #226, how come it was introduced between 1.18.2 and 1.18.3? 1.18.3 does not include any change to the browser field, it already existed before: https://github.com/punkave/sanitize-html/compare/1.18.2...1.18.3
If the tags are not wrong, this must have been caused by a silent dependency update through removal of the lockfile.

dbkr:
Apologies for commenting on closed bugs, but just to resolve the mystery, it turns out adding `"browser": "dist/sanitize-html.js"` to `package.json` does actually cause the build of `sanitize-html.js` to fail in this way, presumably because browserify tries to import the package in the working directory, honouring the `browser` key, so this explains how the bad build happened.

> If the bug was introduced by #226, how come it was introduced between 1.18.2 and 1.18.3? 1.18.3 does not include any change to the browser field, it already existed before: 1.18.2...1.18.3

@felixfbecker see the 2nd problem over in https://github.com/punkave/sanitize-html/issues/241#issuecomment-409881724 - #226 was only actually published to npm in `1.18.3`.

felixfbecker:
Ah, that makes sense, so it was the tags being wrong. Sorry for the confusion.

PR #250, opened 2018-08-30
Stop double encoding ampersands on HTML entities

WillGibson:
Addresses https://github.com/punkave/sanitize-html/issues/249

### DIFF
--- a/README.md
+++ b/README.md
@@ -480,6 +480,10 @@ The content still gets escaped properly, with the exception of the `script` and 
 
 ## Changelog
 
+2.0.0:
+
+* Stop ampersands on HTML entities when parser.decodeEntities = false
+
 1.18.4:
 
 * Removed incorrect `browser` key, restoring frontend build. Thanks to Felix Becker.
--- a/README.md
+++ b/README.md
@@ -480,9 +480,9 @@ The content still gets escaped properly, with the exception of the `script` and 
 
 ## Changelog
 
-2.0.0:
+1.18.5:
 
-* Stop ampersands on HTML entities when parser.decodeEntities = false
+* Stop double encoding ampersands on HTML entities
 
 1.18.4:
 
--- a/test/test.js
+++ b/test/test.js
@@ -751,13 +751,14 @@ describe('sanitizeHtml', function() {
   it('Should encode &, <, > and "', function() {
     assert.equal(sanitizeHtml('"< & >"'), '&quot;&lt; &amp; &gt;&quot;');
   });
-  it('Should not encode ampersands on HTML entities when parser.decodeEntities = false', function() {
-    var text = 'This &amp; that &reg';
+  it('Should not double encode ampersands on HTML entities', function() {
+    var textIn = 'This &amp; & that &reg';
+    var expectedResult = 'This &amp; &amp; that &reg';
     var sanitizeHtmlOptions = {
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -480,6 +480,10 @@ The content still gets escaped properly, with the exception of the `script` and
 
 ## Changelog
 
+2.0.0:
```
WillGibson: @boutell Technically this PR breaks backwards compatibility, so we should bump the major version in case anyone is relying on the behaviour I've changed.

boutell:
Can't this be used to sneak in entity escapes when typing in a rich text editor, for instance? That doesn't sound like desirable behavior.

boutell:
I'm not sure, however, where the bc break is if it's optional.

test/test.js
```diff
@@ -748,4 +748,16 @@ describe('sanitizeHtml', function() {
         allowedSchemes: sanitizeHtml.defaults.allowedSchemes.concat([ 'tel' ]),
       }), '<q cite=\"http://www.google.com\">HTTP</q><q cite=\"https://www.google.com\">HTTPS</q><q cite=\"mailto://www.google.com\">MAILTO</q><q cite=\"tel://www.google.com\">TEL</q><q cite=\"ftp://www.google.com\">FTP</q><q>DATA</q><q>LDAP</q><q>ACROBAT</q><q>VBSCRIPT</q><q>FILE</q><q>RLOGIN</q><q>WEBCAL</q><q>JAVASCRIPT</q><q>MMS</q>');
   });
+  it('Should encode &, <, > and "', function() {
+    assert.equal(sanitizeHtml('"< & >"'), '&quot;&lt; &amp; &gt;&quot;');
+  });
+  it('Should not encode ampersands on HTML entities when parser.decodeEntities = false', function() {
```
WillGibson: Re: _Can't this be used to sneak in entity escapes when typing in a rich text editor, for instance? That doesn't sound like desirable behavior._

I'm not sure about that, but converting things like `&reg; to &amp;reg can't be right. Maybe the test I wrote is wrong, and what it actually needs to do it simply none double encode things?

README.md
WillGibson: Re: _I'm not sure, however, where the bc break is if it's optional._

If someone was already using `parser.decodeEntities = false`, and what they get back from `sanitiseHtml()` changed from what they were expecting, I would say that breaks backwards compatibility.

E.g. this which came to light about an hour after my very first open source PR was merged despite me having mention possible BC breaking...

https://github.com/Behat/Transliterator/pull/18#commitcomment-21637919

test/test.js
WillGibson: @boutell Thanks for for feedback before. I've now updated the PR and I hope it might be acceptable now.

WillGibson:
@boutell Is the updated PR OK?

boutell:
Merged, thanks.

WillGibson:
Thank you @boutell

WillGibson:
@boutell Is it you I need to ask nicely to tag the 1.18.5 release and publish to NPM?

boutell:
Yep. Published. (You might not see it right away, npm is lagging behind lately in showing package updates to the world.)

WillGibson:
That's great, thanks @boutell

PR #252, opened 2018-09-12
Add allowIframeRelativeUrls option

rickpmartin:
The option is pretty self-explanatory, the only real complication is in maintaining the old behavior when no option value is specified.  I added both a comment in the code and a note in the readme explaining that the implicit value (meaning the value when no value set specifically via options) is:

false when allowIframeHostnames specified
true otherwise


boutell:
Thanks!

PR #262, opened 2018-10-29
Update default options in README.md

uturnr:
Updated default options in README.md to match sanitizeHtml.defaults in index.js


boutell:
Thanks.

PR #264, opened 2018-11-16
Upgrade dependencies

jbraithwaite:


boutell:
You bumped several modules to new major versions. Have you reviewed for bc breaks that could impact use?

jbraithwaite:
The existing tests passed, is there anything you'd like me to manually tests? Or more tests you'd like written?

boutell:
(I don't see any major version changes to modules whose APIs we re-export without restriction via an option, so as long as the tests pass!)

PR #267, opened 2018-11-29
Remove dependency on postcss

jbraithwaite:
`postcss` is a pretty heavy weight dependency if it is only being used for parsing an AST.

Replaced with [css-tree](https://www.npmjs.com/package/css-tree).

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -6,7 +6,7 @@ var mergeWith = require('lodash.mergewith');
 var isString = require('lodash.isstring');
 var isPlainObject = require('lodash.isplainobject');
 var srcset = require('srcset');
-var postcss = require('postcss');
+var csstree = require('css-tree');
 var url = require('url');
 
 function each(obj, cb) {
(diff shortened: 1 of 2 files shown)


src/index.js
```diff
@@ -299,15 +299,31 @@ function sanitizeHtml(html, options, _recursing) {
...
+                    var rules = selectors[node.property];
+
+                    if (rules !== undefined && rules.every(function (rule) { return !value.match(rule); })) {
+                      list.remove(item)
+                    }
+                  }
+                })
+
+                value = csstree.generate(ast).slice(name.length + 1);
+                value = value.slice(0, value.length - 1);
```
squirly: You may want to extract 302-317 to a function.

jbraithwaite:
From their readme:

> NOTE: The project is in alpha stage since some parts need further improvements, AST format and API are subjects to change. However it's stable enough and used by packages like CSSO (CSS minifier) and SVGO (SVG optimizer) in production.

npm reports 8 million downloads a month for this package and it is depended on [by these packages](https://www.npmjs.com/browse/depended/css-tree) (including SVGO and CSSO).

jbraithwaite:
I can write additional tests if you think that's required.

boutell:
This has been rolled back for now (1.19.3), pending further examination of #269, please take a look. Thanks!

PR #280, opened 2019-02-05
Removed toLowerCase from naughtyHref

subhashdasyam:
Removed toLowerCase from naughtyHref to enable CAPITAL , lower and MiXed allowedSchemes are allowed. 
This was one of the blocker for me.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -456,7 +456,7 @@ function sanitizeHtml(html, options, _recursing) {
       // No scheme
       return false;
     }
-    var scheme = matches[1].toLowerCase();
+    var scheme = matches[1];
 
     if (has(options.allowedSchemesByTag, name)) {
       return options.allowedSchemesByTag[name].indexOf(scheme) === -1;


boutell:
How was this a blocker for you? Chrome treats the scheme as case insensitive. We would miss that and reject it if we did not do this.

subhashdasyam:
> How was this a blocker for you? Chrome treats the scheme as case insensitive. We would miss that and reject it if we did not do this.

The problem is for our group code we have several custom protocols in a href tag like 
CUST1://
CuSt2://
company1://
ComPany2://

instead of http or https:// for mobile applications so that users when clicked on links will redirect to their own respective APPS. 

I hope i made sense.

Regards

abea:
@boutell Do you think there's reason to carry this forward?

boutell:
It's not pressing, but if the OP is interested in devising a solution where we test case insensitively for all protocols not explicitly marked as case sensitive, we could look at that.

PR #299, opened 2019-07-26
Upgrade: upgrade lodash.mergewith": "^4.6.2" for security (fix #298)

### LINKED ISSUE #298: Upgrade: lodash.merge security vulnerability at 4.6.1
- https://github.com/apostrophecms/sanitize-html/blob/master/package.json#L37 lists the vulnerable lodash.merge version
- Per https://github.com/lodash/lodash/issues/4348, an upgrade is required to address a recent security vulnerability CVE-2019-10744

Will follow up with a PR for this today or tomorrow if others do not get to it / other feedback does not come in.

quetzaluz:
- See issue #298


boutell:
Due to the use of semver ranges this is not a security issue, there's no package-lock for the module and even if there were installation of a larger project that depends on it would only look at its own package-lock. But, it doesn't hurt to tick the version number up.

Yohanna:
Any updates on this PR? I'm hitting this security issue in my security scans.

abea:
@Yohanna Sorry about the delay. We'll get this deployed tomorrow. You should be able to remove `lodash.mergewith` from your project's lock file, delete the node_modules directory and reinstall to already get the newer version, however.

boutell:
Because of semantic versioning the exact version mentioned in package.json should not matter at all, just typing `npm update` should fix your project. But, I think I've seen some flakiness with that which is why we've been putting out a few releases solely to "make it official" that only newer versions of dependencies are OK, even though the versioning rules in place are fine.

Yohanna:
Perfect, thanks everyone! :)

PR #310, opened 2019-10-26
Fix tests

jbraithwaite:
Tests were previously not running because the `npm` script was running `prepublishOnly` which was not producing `/dist/index.js`.

Now `npm run test` runs `build` which produces `/dist/sanitize-html.js` that mocha can use.


boutell:
The point of prepublishOnly was to test specifically what will happen on publish.

If it's not running then, I'm not sure it's running at npm publish time at all; can you shed any light on that?

jbraithwaite:
Typically, you'd want to be testing during development (locally and in CI) not when you are about publish the package.

boutell:
Having the tests pass as part of "npm publish" is part of our release practice. Any changes that break that we'll need adjustment.

abea:
It looks like this was resolved a few days after this PR was created in #311

PR #311, opened 2019-10-29
Fixed require statement in test.js

yehonatanz:
`npm test` is currently broken, this PR fixes it.
The problem was `require`-ing the wrong filename in `test.js` (`index.js` instead of `sanitize-html.js`).


boutell:
Thanks.

PR #324, opened 2020-01-19
Update srcset dependency from v1 to v2

simison:
Seems mostly meta release and drops one dependency: https://github.com/sindresorhus/srcset/releases

Tested that `npm test` passes.


boutell:
Thanks.

Issue #352, opened 2020-05-14
exclusiveFilter doesn't follow disallowedTagsMode=escape


thiago-negri:
exclusiveFilter is always removing the tag from the text instead of escaping when disallowedTagsMode=escape

```javascript
test('exclusiveFilter with escape', () => {
  const config = {
    disallowedTagsMode: 'escape',
    allowedTags: [ 'a' ],
    allowedAttributes: { 'a': [ 'href' ] },
    allowedSchemes: [ 'http', 'https' ],
    exclusiveFilter(frame): boolean {
        return (frame.tag === 'a' && /google.com/.test(frame.attribs.href));
    },
  };
  const input = '<a href="ok.com">ok</a> <a href="google.com">google</a> <a href="other.com">other</a>';
  const expected = '<a href="ok.com">ok</a> &lt;a href="google.com"&gt;google&lt;/a&gt; <a href="other.com">other</a>';
  const output = sanitizeHtml(input, config);

  // fails, actual output: '<a href="ok.com">ok</a>  <a href="other.com">other</a>'
  expect(output).to.equal(expected);
});
```

abea:
Thanks for the contribution, but I don't think this is a bug. The description of the option is "You can provide a filter function to _remove_ unwanted tags" (my emphasis). The intention is to remove them. There's a valid use case for exclusive escaping, but that would be an enhancement.

PR #355, opened 2020-05-14
Add test for exclusiveFilter escape bug

thiago-negri:
Test for https://github.com/apostrophecms/sanitize-html/issues/352

It's failing


abea:
As noted in #352 , I think this is testing for an unintended use case.

PR #360, opened 2020-05-25
Introduce allowedIframeDomains

### LINKED ISSUE #359: Allow any subdomain of a host in allowedIframeHostnames
Hello!
I need any subdomain of `zoom.us` to be allowed for embedding.
Now I allowed `us02web.zoom.us` but I was told that any other subdomain may appear in the future.
I think the most flexible solution is to process RegEx alongside String.

StanisLove:
resolve #359

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -279,9 +279,6 @@ function sanitizeHtml(html, options, _recursing) {
                 // naughtyHref is in charge of whether protocol relative URLs
                 // are cool. We should just accept them
                 parsed = url.parse(value, false, true);
-                parsed.domain = parsed.hostname && parsed.hostname.substring(
-                  parsed.hostname.lastIndexOf('.', parsed.hostname.lastIndexOf('.') - 1) + 1
-                );
                 var isRelativeUrl = parsed && parsed.host === null && parsed.protocol === null;
                 if (isRelativeUrl) {
                   // default value of allowIframeRelativeUrls is true
@@ -294,7 +291,7 @@ function sanitizeHtml(html, options, _recursing) {
                     return hostname === parsed.hostname;
                   });
                   var allowedDomain = (options.allowedIframeDomains || []).find(function(domain) {
-                    return domain === parsed.domain;
+                    return parsed.hostname.endsWith(domain);
                   });
                   allowed = allowedHostname || allowedDomain;
                 }
--- a/src/index.js
+++ b/src/index.js
@@ -279,15 +279,24 @@ function sanitizeHtml(html, options, _recursing) {
                 // naughtyHref is in charge of whether protocol relative URLs
                 // are cool. We should just accept them
                 parsed = url.parse(value, false, true);
+                parsed.domain = parsed.hostname && parsed.hostname.substring(
+                  parsed.hostname.lastIndexOf('.', parsed.hostname.lastIndexOf('.') - 1) + 1
+                );
                 var isRelativeUrl = parsed && parsed.host === null && parsed.protocol === null;
                 if (isRelativeUrl) {
-                  // default value of allowIframeRelativeUrls is true unless allowIframeHostnames specified
+                  // default value of allowIframeRelativeUrls is true
+                  // unless allowedIframeHostnames or allowedIframeDomains specified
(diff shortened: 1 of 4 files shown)


StanisLove:
@boutell Thanks for reviewing. I have applied your suggestion to introduce allowedIframeDomains option.

src/index.js
```diff
@@ -279,15 +279,24 @@ function sanitizeHtml(html, options, _recursing) {
                 // naughtyHref is in charge of whether protocol relative URLs
                 // are cool. We should just accept them
                 parsed = url.parse(value, false, true);
+                parsed.domain = parsed.hostname && parsed.hostname.substring(
+                  parsed.hostname.lastIndexOf('.', parsed.hostname.lastIndexOf('.') - 1) + 1
```
boutell, line 283: If example.com is an allowed domain, this approach will match foo.example.com, but not foo.2.example.com.

A better approach would be to check whether the hostname either exactly matches the domain, or ends in a period followed by the domain. Then the number of dots in the prefix is not relevant.

StanisLove: It really works.
image
But I agree that it isn't a clear piece of code and it is better to use `endsWith` checking.

StanisLove:
@boutell Let me know that it's ok and I will update the documentation.

src/index.js
```diff
@@ -281,13 +281,19 @@ function sanitizeHtml(html, options, _recursing) {
...
-                } else if (options.allowedIframeHostnames) {
-                  allowed = options.allowedIframeHostnames.find(function (hostname) {
+                    ? options.allowIframeRelativeUrls
+                    : (!options.allowedIframeHostnames && !options.allowedIframeDomains);
+                } else if (options.allowedIframeHostnames || options.allowedIframeDomains) {
+                  var allowedHostname = (options.allowedIframeHostnames || []).find(function (hostname) {
                     return hostname === parsed.hostname;
                   });
+                  var allowedDomain = (options.allowedIframeDomains || []).find(function(domain) {
+                    return parsed.hostname.endsWith(domain);
```
boutell, line 294: Need to make sure it's either matching the entire string, or ending with the domain preceded by a dot. Otherwise you can fake it out with a newly registered domain name, for instance if youtube.com is allowed you can register fakeyoutube.com and it will pass.

boutell: I appreciate the effort, sorry for the nitpicking but it we do need to secure it.

StanisLove: Oh, you are deadly right!

StanisLove: I'll cover this case with a test.

StanisLove:
@boutell I have taken account of your concerns.
Also I have taken up the case when a top-level domain can be passed to whitelisted domains. (F.e. `['com']`)
I think it wouldn't be the desired behavior to allow all sources in the `.com` domain zone, so I have added the [guard clause](https://github.com/apostrophecms/sanitize-html/pull/360/files#diff-1fdf421c05c1140f6d71444ea2b27638R295-R297).

src/index.js
```diff
@@ -281,13 +281,24 @@ function sanitizeHtml(html, options, _recursing) {
...
-                } else if (options.allowedIframeHostnames) {
-                  allowed = options.allowedIframeHostnames.find(function (hostname) {
+                    ? options.allowIframeRelativeUrls
+                    : (!options.allowedIframeHostnames && !options.allowedIframeDomains);
+                } else if (options.allowedIframeHostnames || options.allowedIframeDomains) {
+                  var allowedHostname = (options.allowedIframeHostnames || []).find(function (hostname) {
                     return hostname === parsed.hostname;
                   });
+                  var allowedDomain = (options.allowedIframeDomains || []).find(function(domain) {
+                    // avoid top-level domains
```
boutell, line 294: One last annoying request: please remove the guard. `allowedIframeDomains` is not user input, and I don't see someone putting 'com' in there by accident; we should allow developers to do what they want to do. Consider the case where a custom top level domain is valid only on a certain intranet; here it could make sense.

boutell: I appreciate the intent though.

boutell:
Note to @abea we need to make sure this gets changelogged and announced.

PR #361, opened 2020-05-26
Enforce html boundary

### LINKED ISSUE #149: Strip characters after </html>

### LINKED ISSUE #164: garbage text not cleaned up

abea:
Revives the work started in #166 

Resolves #164
Resolves #149


README.md
```diff
@@ -187,7 +188,15 @@ allowedAttributes: {
   '*': [ 'href', 'align', 'alt', 'center', 'bgcolor' ]
 }
 ```
+### Discarding text outside of ```<html></html>``` tags
 
+Sometimes various applications creating html as function of clipboard copy generate control characters after terminating html tag. By default sanitize-html will not discard these characters and return them in sanitized string. This behaviour can be modified using enforceHtmlBoundary option.
```
boutell, line 193: This language needs some cleanup.

src/index.js
```diff
@@ -159,15 +159,33 @@ function sanitizeHtml(html, options, _recursing) {
...
-  var skipText = false;
-  var skipTextDepth = 0;
+  var depth;
+  var stack;
+  var skipMap;
+  var transformMap;
+  var skipText;
+  var skipTextDepth;
+
+  var initializeState = function() {
```
boutell, line 169: Follow style of the module, use a named function and place it lower down, top-down programming.

PR #362, opened 2020-05-28
Adds option to the nonTextTagsArray array

### LINKED ISSUE #60: Strip content / text, not just tags
Currently when a tag is not allowed this results in a lot of (strange) text. E.g. a select box of country options would result in a long sentence of country names, if the option element is not allowed.

Is it possible to delete these tags entirely, including the contents / text?

abea:
Closes #60

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,9 @@
 
 `sanitize-html` allows you to specify the tags you want to permit, and the permitted attributes for each of those tags.
 
-If a tag is not permitted, the contents of the tag are still kept, except for `script`, `style` and `textarea` tags.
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -10,7 +10,9 @@
 
 `sanitize-html` allows you to specify the tags you want to permit, and the permitted attributes for each of those tags.
 
-If a tag is not permitted, the contents of the tag are still kept, except for `script`, `style` and `textarea` tags.
+If a tag is not permitted, the contents of the tag are not discarded. There are
```
abea, line 13: I wanted to avoid any future mistakes from not updating all the locations.

PR #365, opened 2020-06-02
Replace xtend with builtin Object.assign

TrySound:
Ref https://github.com/apostrophecms/sanitize-html/issues/257

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,6 +1,5 @@
 /* eslint-disable no-useless-escape */
 var htmlparser = require('htmlparser2');
-var extend = require('xtend');
 var quoteRegexp = require('lodash/escapeRegExp');
 var cloneDeep = require('lodash/cloneDeep');
 var mergeWith = require('lodash/mergeWith');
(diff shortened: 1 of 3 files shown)


src/index.js
```diff
@@ -101,9 +100,9 @@ function sanitizeHtml(html, options, _recursing) {
     options = sanitizeHtml.defaults;
     options.parser = htmlParserDefaults;
   } else {
-    options = extend(sanitizeHtml.defaults, options);
+    options = Object.assign({}, sanitizeHtml.defaults, options);
```
abea, line 103: I think we can let `sanitizeHtml.defaults` be the target argument since it's a set of defaults.

src/index.js
```diff
@@ -101,9 +100,9 @@ function sanitizeHtml(html, options, _recursing) {
...
     if (options.parser) {
-      options.parser = extend(htmlParserDefaults, options.parser);
+      options.parser = Object.assign({}, htmlParserDefaults, options.parser);
```
abea, line 105: Same here, making `htmlParserDefaults` the target argument.

TrySound:
Mutating defaults does not look right. I used Object.assign in immutable way similar to default xtend implementation.
https://github.com/Raynos/xtend/blob/master/package.json#L15

TrySound:
This is especially wrong because you expose defaults to the user.

TrySound:
Object.assign mutates target object by the way.

TrySound:
@abea Rebased and fixed base branch.

PR #368, opened 2020-06-03
Fix srcset parsing

bard:
Replace [srcset](https://github.com/sindresorhus/srcset) with [parse-srcset](https://github.com/albell/parse-srcset) to address https://github.com/apostrophecms/sanitize-html/issues/367.

Additionally, fix tests with invalid srcset containing both width descriptor and pixel density descriptor. [Only zero or one of those is allowed](https://html.spec.whatwg.org/multipage/images.html#srcset-attributes).

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,8 +29,8 @@
     "chalk": "^2.4.1",
     "htmlparser2": "^4.1.0",
     "lodash": "^4.17.15",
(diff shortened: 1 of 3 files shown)


abea:
There are also some more tests with both the width and pixel density properties. If you want to fix those as well we'd appreciate it. Maybe leave some with pixel density instead of width.

Issue #136, opened 2016-12-18
transformTags return double text when changing text of tag that already have text


newz:
Example.
```
var sanitizeHtml = require("sanitize-html");

var clean = sanitizeHtml('Go to <a href="https://www.google.com">Google</a>', {
  transformTags: {
    'a': function(tagName, attribs) {
        return {
            tagName: 'a',
            text: 'Github',
            attribs: {
                href: "https://www.github.com"
            }
        };
    }
  }
});

console.log(clean);
```
will result in
```
Go to <a href="https://www.github.com">GithubGithub</a>
```

You can see that transformTags return Github twice.

Runkit url
https://runkit.com/5856b06df18eae0013bcf2da/5856b974001b65001374e127

halffullheart:
Experiencing the same issue.

george-norris-salesforce:
Me too

civalin:
me too

# example

```javascript
var sanitizeHtml = require('sanitize-html');
sanitizeHtml('x<a>y</a>', { transformTags: { a: () => { return { tagName: 'div', text: 'z' }; } }});
```

the result is:

```
'x<div>zz</div>'
```

(expect `'x<div>z</div>'`)

# version info

- sanitize-html@1.14.1
- node.js 6.10.2

andpor:
This is definitely a defect. I can reproduce it easily with 1.14.1. Actually the unit test case will fail as well. This works fine if returned text is empty string or a single space: '' or ' ' - duplicate does not occur for these values.

andpor:
This is unit test case. Currently it is not failing because text returned from transformTags is empty string, but if it is slightly changed to return value "google.com" the output is duplicated.

`
```
it('should replace text and attributes when they are changed by transforming function', function () {
    assert.equal(sanitizeHtml('<a href="http://somelink">some text</a>', { transformTags: {a: function (tagName, attribs) {
      return {
        tagName: tagName,
        attribs: attribs,
        text: 'google.com'
      }
    }}}), '<a href="http://somelink">google.com</a>');
  });
`

```
65 passing (12s)
  1 failing

  1) sanitizeHtml should replace text and attributes when they are changed by transforming function:

      AssertionError: '<a href="http://somelink">google.comgoogle.com</a>' == '<a href="http://somelink">google.com</a>'
      + expected - actual

      -<a href="http://somelink">google.comgoogle.com</a>
      +<a href="http://somelink">google.com</a>
      
[...]

Issue #125, opened 2016-08-23
duplicate output when element has just &nbsp;


omerts:
I am trying to transform a tag that only has `&nbsp` as its content, and the result is a duplicate of the transformation.

Reproduction:

``` javascript
sanitize('<span>&nbsp</span>', {
          transformTags: {
            'span': (tag, attribs) => {
              return {
                tagName: 'i',
                text: 'Why am I here twice?'
              }
            }
          }
        })
```

Results in:

```
Why am I here twice?Why am I here twice?
```

anyx:
I have same issue. After quick research i found this:

https://github.com/punkave/sanitize-html/blob/master/index.js#L194

```
        if (frame.innerText && !hasText && !options.textFilter) {
          result += frame.innerText;
        }
```

So i put empty textFilter in my sanitize configuration:

```
const sanitzeConfig = _.defaults(
    {
        transformTags: {
            'a': function (tagName, attribs) {
                return {
                    tagName: 'span',
                    text: 'Some text'
                };
            }
        },
        textFilter: function(text) {
            return text;
        },
    },
    sanitizeHtml.defaults
);
```

This workaround works for me

abea:
Closing in favor of #136 since that has additional detail.

PR #378, opened 2020-07-08
Fix/125

abea:
Failing test for #125 and #136 

A `text` value in a `transformTags` function is added twice. Once in `onopentag` and again in `ontext`.


abea:
I think at L213 or so we need to be doing something with `skipText` and `skipTextDepth`. I haven't grokked `skipTextDepth` yet, though.

PR #383, opened 2020-07-16
Basic README updates re: no build

abea:
Updates the README to remove build references

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ But, perhaps you'd like to display sanitized HTML immediately in the browser for
 npm install sanitize-html # yarn install sanitize-html
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -34,28 +34,20 @@ HTML comments are not preserved.
...
+* Run npm install and :
 
 ```bash
-npm install
-npm run minify
+npm install sanitize-html # yarn install sanitize-html
 ```
 
-You'll find the minified and unminified versions of sanitize-html (with all its dependencies included) in the dist/ directory.
+The primary change in the 2.x version of sanitize-html is that it no longer includes a build that is ready for browser use. Developers are expected to include sanitize-html in their project builds as they would any other dependency. So while sanitize-html is no longer ready to link to directly in HTML, developers can now more easily process it according to their needs.
```
boutell, line 44: Say the w-word. If people are confused they should be nudged toward webpack.

PR #386, opened 2020-07-25
Upgrade srcset

TrySound:
Ref https://github.com/sindresorhus/srcset/releases/tag/v3.0.0

The new version matches latest spec and requires node v10.


TrySound:
Forgot. Done

PR #393, opened 2020-07-31
2/no iframe

### LINKED ISSUE #122: h1 and h2 not default tags to keep?
This is more of a question than issue. `h1` and `h2` are omitted from the default allowed tags, while `h[3-6]` are allowed. What's the reasoning behind this? Is there something inherently unsafe about `h1` and `h2`?

### LINKED ISSUE #285: iframe is allowed by default, but should it be?
Apologies if this is outside of the goals of the default tags allowed, but I was surprised to see `iframe` in the list. Doesn't that carry an unacceptable amount of risk with it? Thanks!

abea:
Resolves #122 (add h1 and h2, expands defaults)
Resolves #285 (remove iframe)

Also sends `nl` to join its friends in the XML afterlife.


index.js
```diff
@@ -656,9 +656,26 @@ const htmlParserDefaults = {
   decodeEntities: true
 };
 sanitizeHtml.defaults = {
-  allowedTags: ['h3', 'h4', 'h5', 'h6', 'blockquote', 'p', 'a', 'ul', 'ol',
-    'nl', 'li', 'b', 'i', 'strong', 'em', 'strike', 'abbr', 'code', 'hr', 'br', 'div',
-    'table', 'thead', 'caption', 'tbody', 'tr', 'th', 'td', 'pre', 'iframe'],
+  allowedTags: [
+    // Sections derrived from MDN element categories and limited to the more
```
boutell, line 660: typo: derrived

index.js
```diff
@@ -656,9 +656,26 @@ const htmlParserDefaults = {
...
+    // Content sectioning
+    'address', 'article', 'aside', 'footer', 'header',
+    'h1', 'h2', 'h3', 'h4', 'h5', 'h6', 'hgroup',
+    'main', 'nav', 'section',
+    // Text content
+    'blockquote', 'dd', 'div', 'dl', 'dt', 'figcaption', 'figure',
+    'hr', 'li', 'main', 'ol', 'p', 'pre', 'ul',
+    // Inline text semantics
+    'a', 'abbr', 'b', 'bdi', 'bdo', 'br', 'cite', 'code', 'data', 'dfn',
+    'em', 'i', 'kbd', 'mark', 'q',
```
boutell, line 672: Should 'strike' come back?

boutell: Dropping any benign tags that were on the 1.x list could be more hassle for upgraders than it's worth...

abea: It's marked as obsolete since html5. https://developer.mozilla.org/en-US/docs/Web/HTML/Element/strike

boutell: OK fair.

PR #401, opened 2020-08-14
Replaces includes with _.indexOf

### LINKED ISSUE #400: Array/String.includes does not work in IE11
In latest versions of the library you added the `.includes` function in several parts of the code. And since this experimental feature is not supported by IE11, it breaks the app there. Then, if I try to add the [Polyfill](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes) provided by MDN, it starts causing troubles in places where we loop arrays/strings using `for...in`, because it will iterate over the added "includes" option to the array prototype.

abea:
Closes #400

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## Changelog
 
-1.27.4 (2020-08-12):
+1.27.4 (2020-08-26):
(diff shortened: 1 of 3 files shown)


boutell:
Any reason not to use the indexOf method of arrays? Goes back to the dawn times

abea:
Only because I didn't think of it.

PR #402, opened 2020-08-14
[v2] Upgrade klona

TrySound:
Ref https://github.com/lukeed/klona/releases/tag/v2.0.0

The new version got a few new modes and migrated to named export.


abea:
That works. Please update the changelog. This will be for `2.0.0-beta.3`.

PR #409, opened 2020-09-03
use existing has function, less code

boutell:


TrySound (approved):

abea (approved):

abea (changes requested):
It only needs the changelog and version update.

abea (approved):

PR #410, opened 2020-09-09
Use default export for isPlainObject

bertyhell:
Otherwise i get this error:
```
Uncaught TypeError: isPlainObject is not a function
```

image

I'm not sure if this is related to my webpack settings or specifically to sanitize-html

Tested with: 2.0.0-rc.1


bertyhell:
another option would be to add:
```
const isPlainObjectLib = require('is-plain-object');
let isPlainObject = isPlainObjectLib.default ? isPlainObjectLib.default : isPlainObjectLib;
```

TrySound:
Ideally we should add interop to each require like babel and ts do while transpiling.

TrySound:
Another option is to forbid webpack to use module field from is-plain-object package.

abea:
This PR fails the test suite with that same error. `TypeError: isPlainObject is not a function`. It sounds like this is an issue when compiling with webpack. We don't actively support a build with 2.x, but would be happy to accept a PR that supports builds as long as it doesn't cause problems without a build. The corollary to that would probably be that we'd not want a bunch of conditionals thrown in for this either.

Can you provide a failing test for this? Or maybe the PR should be added documentation for people running their own build.

abea (changes requested):
This PR fails the test suite with that same error. `TypeError: isPlainObject is not a function`. Do make sure you run `npm test` locally on any edits.

It sounds like this is an issue when compiling with webpack. We don't actively support a build with 2.x, but would be happy to accept a PR that supports builds as long as it doesn't cause problems without a build. The corollary to that would probably be that we'd not want a bunch of conditionals thrown in for this either.

Can you provide a failing test for this? Or maybe the PR should be added documentation for people running their own build.

TrySound:
I can also release new major with named export instead of default export. This works with klona at least.

TrySound:
This should fix the problem https://github.com/apostrophecms/sanitize-html/pull/411

abea:
Closing in favor of #411

abea:
Thanks for the PR, @bertyhell

PR #411, opened 2020-09-09
Upgrade is-plain-object

TrySound:
Ref https://github.com/jonschlinkert/is-plain-object/releases/tag/v5.0.0


abea:
Nice! I didn't make the connection that you were a maintainer on that.

abea (changes requested):
Please update the changelog and version to `rc.3` and this should be good to go.

TrySound:
@abea Done

abea (approved):

PR #414, opened 2020-09-16
nestingLimit option

boutell:
There is also an eslint formatting update here, but it should be easier to read with github's "ignore whitespace changes" feature.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## 2.0.0-rc.4 (2020-09-23):
+- `nestingLimit` option added.
+
 ## 2.0.0-rc.3 (2020-09-09):
 - Upgrade `is-plain-object` package with named export. Thanks to [Bogdan Chadkin](https://github.com/TrySound) for the contribution.
 
--- a/README.md
+++ b/README.md
@@ -543,6 +543,16 @@ This will transform `<disallowed>content</disallowed>` to `&lt;disallowed&gt;con
 
 Valid values are: `'discard'` (default), `'escape'` (escape the tag) and `'recursiveEscape'` (to escape the tag and all its content).
 
+### Restricting deep nesting
+
+You can limit the depth of HTML tags in the document with the `nestingLimit` option:
+
+```javascript
+nestingLimit: 6
+```
+
+This will prevent the user from nesting tags more than 6 levels deep. Note that the rules for preserving text inside the nested tags are the same in this situation as they would be if the tags were not on the `allowedTags` list.
+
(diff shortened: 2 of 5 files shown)


index.js
```diff
@@ -224,7 +224,7 @@ function sanitizeHtml(html, options, _recursing) {
         }
       }
 
-      if ((options.allowedTags && options.allowedTags.indexOf(name) === -1) || (options.disallowedTagsMode === 'recursiveEscape' && !isEmptyObject(skipMap))) {
+      if ((options.allowedTags && options.allowedTags.indexOf(name) === -1) || (options.disallowedTagsMode === 'recursiveEscape' && !isEmptyObject(skipMap)) || (options.nestingLimit != null && depth >= options.nestingLimit)) {
```
boutell, line 227: This is the important line.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "sanitize-html",
-  "version": "2.0.0-rc.2",
+  "version": "2.0.0-rc.3",
```
boutell, line 3: I'm following the pattern here but don't let me stop you from releasing 2.0.0 final.

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Changelog
 
+## 2.0.0-rc.4 (2020-09-23):
```
abea, line 3: We're ready to go to `2.0.0`, right?

README.md
```diff
@@ -543,6 +543,16 @@ This will transform `<disallowed>content</disallowed>` to `&lt;disallowed&gt;con
...
 
+### Restricting deep nesting
+
+You can limit the depth of HTML tags in the document with the `nestingLimit` option:
+
+```javascript
+nestingLimit: 6
+```
+
+This will prevent the user from nesting tags more than 6 levels deep. Note that the rules for preserving text inside the nested tags are the same in this situation as they would be if the tags were not on the `allowedTags` list.
```
abea, line 554: What does this mean, "prevent the user from nesting tags?" Does it strip deeper tags out?

package.json
abea: Let's go for the 2.0.0. Unless you want that to be its own PR, which works too.

abea (changes requested):
Mostly the README line, but also the version if you want.

CHANGELOG.md
boutell: Yes please, I was just following the pattern, when you are ready to go to 2.0.0 go ahead and change.

boutell:
2.0.0 should probably be its own PR.

abea (approved):

PR #422, opened 2020-09-29
Allow superscript tag by default

julianlam:
It seems subscript (`<sub>`) is allowed, but `<sup>` is not, so this commit adds `sup` to the list of default allowedTags

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## 2.1.0 (2020-09-30):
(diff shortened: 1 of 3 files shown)


abea (commented):
That looks like a good addition. Please include it in the README documentation of this default and add a Changelog entry for it (a new minor version).

julianlam:
Done and done, thanks! I was hesitant to update the changelog since I wasn't sure whether it was auto-generated.

abea:
Excellent. We are not that fancy.

abea (approved):

abea:
fyi - the next release will be Oct 7th. I'll make the change, though.

PR #429, opened 2020-10-09
Upgrade htmlparser2

TrySound:
Ref https://github.com/fb55/htmlparser2/releases/tag/v5.0.0

### DIFF
--- a/index.js
+++ b/index.js
@@ -109,17 +109,9 @@ function sanitizeHtml(html, options, _recursing) {
     };
   }
 
-  if (!options) {
-    options = sanitizeHtml.defaults;
-    options.parser = htmlParserDefaults;
-  } else {
-    options = Object.assign({}, sanitizeHtml.defaults, options);
-    if (options.parser) {
-      options.parser = Object.assign({}, htmlParserDefaults, options.parser);
-    } else {
-      options.parser = htmlParserDefaults;
-    }
-  }
+  options = Object.assign({}, sanitizeHtml.defaults, options);
+  options.parser = Object.assign({}, htmlParserDefaults, options.parser);
+
   // vulnerableTags
   vulnerableTags.forEach(function (tag) {
     if (
--- a/test/test.js
+++ b/test/test.js
@@ -681,7 +681,7 @@ describe('sanitizeHtml', function() {
   it('should not crash due to tag names that are properties of the universal Object prototype', function() {
     assert.equal(
       sanitizeHtml('!<__proto__>!'),
-      '!!');
+      '!&lt;__proto__&gt;!');
   });
   it('should correctly maintain escaping when allowing a nonTextTags tag other than script or style', function() {
     assert.equal(
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,10 @@
 # Changelog
 
+## 2.3.0 (2020-12-16):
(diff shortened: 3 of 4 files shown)


index.js
```diff
@@ -109,17 +109,9 @@ function sanitizeHtml(html, options, _recursing) {
...
-  } else {
-    options = Object.assign({}, sanitizeHtml.defaults, options);
-    if (options.parser) {
-      options.parser = Object.assign({}, htmlParserDefaults, options.parser);
-    } else {
-      options.parser = htmlParserDefaults;
-    }
-  }
+  options = Object.assign({}, sanitizeHtml.defaults, options);
+  options.parser = Object.assign({}, htmlParserDefaults, options.parser);
```
TrySound, line 113: Object.assign ignores anything not object which allows to simplify options extension.

TrySound:
@abea Is this good to go?

abea:
Sorry for the delay @TrySound. I'll review this early this week.

test/test.js
```diff
@@ -681,7 +681,7 @@ describe('sanitizeHtml', function() {
   it('should not crash due to tag names that are properties of the universal Object prototype', function() {
     assert.equal(
       sanitizeHtml('!<__proto__>!'),
-      '!!');
+      '!&lt;__proto__&gt;!');
```
abea, line 684: @boutell Do you have thoughts about this resolution? It's still not crashing, as is the goal of the test, but the output is different. I'm not totally clear on why the tag is being escaped, regardless.

@TrySound Do you know why this is being escaped instead of discarded?

abea:
@TrySound Thanks for your patience with us on this. It's been a busy time.

test/test.js
TrySound: I don't remember to be honest. Will check

TrySound: @abea Alright, I figured it out. Previously `<__proto__>` was parsed as html tag. In this commit valid characters for html element start were restricted to a-zA-Z. So here proto "tag" is escaped as text.
https://github.com/fb55/htmlparser2/pull/497

boutell: According to this, escaping is a sensible response:

https://stackoverflow.com/questions/22066705/invalid-tags-in-html-abc-vs-1234

I don't think we're introducing a bc break here because tags not starting with a letter are not compliant anyway, and even in earlier versions of HTML they got treated as text, not tags. It's OK.

boutell: However a bigger issue is that we allow options to be passed directly to htmlparser2 when calling sanitize-html.

So moving to 6.x is a bc break requiring a major version bump of our own — if that option set has changed in a non-backwards-compatible way. If it has just been extended in a backwards compatible way it's fine, but the htmlparser2 docs have to be checked on this point before acceptance.

boutell (changes requested):

test/test.js
TrySound: From what I see here https://github.com/fb55/htmlparser2/releases/tag/v5.0.0 only decodeEntities is changed but it was set by this project anyway.

v6 changes also do no break sanitize-html https://github.com/fb55/htmlparser2/releases/tag/v6.0.0

abea:
That sounds reasonable. https://github.com/apostrophecms/sanitize-html/pull/429#discussion_r543485923

boutell (approved):

PR #432, opened 2020-10-13
Wildcard support for allowedClasses

### LINKED ISSUE #253: Setting allwedClasses for "*" tag allows all classes, regardless of the filter
```js
sanitize('<span class="not-allowed">Text</span>', {
        allowedTags: [...sanitize.defaults.allowedTags, 'span'],
        allowedClasses: {
            '*': ['allowed'],
        },
    })
```
#### Expected:
```html
<span>Text</span>
```
#### Actual:
```html
<span class="not-allowed"></span>
```
This does not happen when `span` instead of `'*'` is used as the tag name for `allowedClasses`.

Cortys:
This should fix #253.

### DIFF
--- a/README.md
+++ b/README.md
@@ -218,6 +218,13 @@ const clean = sanitizeHtml(dirty, {
 });
 ```
 
+Similar to `allowedAttributes`, you can use `*` as a tag name, to allow listed classes to be valid for any tag:
+```js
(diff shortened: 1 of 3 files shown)


abea (approved):

abea:
Thanks for this contribution!

PR #439, opened 2020-11-02
Fix small things in README

elefevre:
- format all mentions of sanitize-html in the same way
- fix a link
- add a few commas for readability
- fix the case of some characters
- remove space before colons
- align a response to a FAQ

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,11 +4,11 @@
 
 <a href="https://apostrophecms.com/"><img src="https://raw.githubusercontent.com/apostrophecms/sanitize-html/main/logos/logo-box-madefor.png" align="right" /></a>
 
-`sanitize-html` provides a simple HTML sanitizer with a clear API.
+sanitize-html provides a simple HTML sanitizer with a clear API.
 
-`sanitize-html` is tolerant. It is well suited for cleaning up HTML fragments such as those created by CKEditor and other rich text editors. It is especially handy for removing unwanted CSS when copying and pasting from Word.
+sanitize-html is tolerant. It is well suited for cleaning up HTML fragments such as those created by CKEditor and other rich text editors. It is especially handy for removing unwanted CSS when copying and pasting from Word.
 
-`sanitize-html` allows you to specify the tags you want to permit, and the permitted attributes for each of those tags.
+sanitize-html allows you to specify the tags you want to permit, and the permitted attributes for each of those tags.
 
 If a tag is not permitted, the contents of the tag are not discarded. There are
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -35,15 +35,15 @@ HTML comments are not preserved.
...
 * Clone repository and install via npm
-* Run npm install and :
+* Run npm install and:
 
 ```bash
 npm install sanitize-html # yarn install sanitize-html
 ```
 
-The primary change in the 2.x version of sanitize-html is that it no longer includes a build that is ready for browser use. Developers are expected to include sanitize-html in their project builds (e.g., webpack) as they would any other dependency. So while sanitize-html is no longer ready to link to directly in HTML, developers can now more easily process it according to their needs.
+The primary change in the 2.x version of `sanitize-html` is that it no longer includes a build that is ready for browser use. Developers are expected to include `sanitize-html` in their project builds (e.g., webpack) as they would any other dependency. So while `sanitize-html` is no longer ready to link to directly in HTML, developers can now more easily process it according to their needs.
```
abea, line 44: These are the name of the package, which is "sanitize-html" (for better or worse). So I wouldn't format them as code.

README.md
```diff
@@ -258,17 +260,17 @@ const clean = sanitizeHtml(dirty, {
 
 ### Discarding text outside of ```<html></html>``` tags
 
-Some text editing applications generate HTML to allow copying over to a web application. These can sometimes include undesireable control characters after terminating `html` tag. By default sanitize-html will not discard these characters, instead returning them in sanitized string. This behaviour can be modified using `enforceHtmlBoundary` option.
+Some text editing applications generate HTML to allow copying over to a web application. These can sometimes include undesireable control characters after terminating `html` tag. By default `sanitize-html` will not discard these characters, instead returning them in sanitized string. This behaviour can be modified using `enforceHtmlBoundary` option.
```
abea, line 263: Same as above regarding name formatting. Maybe we should make it clearer, naming the package "Sanitize HTML" or something, but that's not the case right now.

abea (changes requested):
Thank you, @elefevre! See my note about the name of the package. I realize the first section does use the code syntax. Since those aren't code instructions, let's make them all plain strings, lowercase even at the beginning of a sentence. It would be worth us reconsidering a non-code title/label for the package.

README.md
elefevre: Alright, I'm going to rename them all to sanitize-html, w/o the backquotes.

abea (approved):

abea:
Thank you!

PR #444, opened 2020-11-26
Allow 'tel' links by default

### LINKED ISSUE #443: Allow href to contain 'tel:'
## The problem to solve

Href's can not (by default as expected from docs) contain a tel: link. Would be nice to have this as default.

## Proposed solution

Put it in the defaults

aHerbots:
fixes #443


abea (changes requested):
This looks good to me. Thank you! Please add a changelog entry and update the README defaults.

PR #456, opened 2021-01-22
Change use of Array `includes` to `indexOf` for IE11 compatibility

craigkovatch:
The use of `includes` in `updateParentNodeMediaChildren` breaks compatibility with IE11, as it was not introduced until ES2016. This simple change uses the perf-equivalent `indexOf` method instead.

### DIFF
--- a/index.js
+++ b/index.js
@@ -102,7 +102,7 @@ function sanitizeHtml(html, options, _recursing) {
     };
 
     this.updateParentNodeMediaChildren = function() {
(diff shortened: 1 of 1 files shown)


abea:
Please see the "browser" section of the README. This isn't meant to be browser-ready, unlike 1.x, which had a backwards compatible build included. It should be included in your front end build process or be compiled as needed for your project. https://github.com/apostrophecms/sanitize-html#browser

PR #469, opened 2021-04-07
Add useAutoClosingTag option

potter7050:
### Summary
- I suggest the `useAutoClosingTag` option to fix the git issue below.

### Reference
- https://github.com/apostrophecms/sanitize-html/issues/464

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,7 +126,8 @@ allowedSchemes: [ 'http', 'https', 'ftp', 'mailto', 'tel' ],
 allowedSchemesByTag: {},
 allowedSchemesAppliedToAttributes: [ 'href', 'src', 'cite' ],
 allowProtocolRelative: true,
-enforceHtmlBoundary: false
+enforceHtmlBoundary: false,
+useAutoClosingTag: true
 ```
 
 ### Common use cases
--- a/index.js
+++ b/index.js
@@ -511,7 +511,9 @@ function sanitizeHtml(html, options, _recursing) {
         return;
       }
 
-      result += '</' + name + '>';
+      if (options.useAutoClosingTag) {
+        result += '</' + name + '>';
+      }
       if (skip) {
         result = tempResult + escapeHtml(result);
         tempResult = '';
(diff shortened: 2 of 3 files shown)


abea:
@potter7050 Thanks for the contribution. On a style note, I would suggest calling this option `autoCloseTags`.

The bigger issue is that setting `useAutoCloseTag: false` does not only disabling closing unclosed tags. It also fails to close tags that are _already closed_ in the markup. See this test:

```javascript
    const wackyText = sanitizeHtml('<p>here\'s a string with a <wacky> tag.</p>', {
      useAutoClosingTag: false,
      disallowedTagsMode: 'escape'
    });
    assert.equal(wackyText, '<p>here\'s a string with a &lt;wacky&gt; tag.</p>');
```

This fails with output `<p>here's a string with a &lt;wacky&gt; tag.`.

Here we have a closed paragraph tag with an unclosed fake tag. The fake tag isn't closed, but the paragraph tag isn't closed either. If we were to call this `closeTags: false`, this would no longer be serving to sanitize HTML.

A fix for the original problem would need to 
1. Recognize that a tag is not recognized (disallowed). If recognized, there should be no change.
2. Identify if the unrecognized tag is already closed in the input. If it is already closed, there should be no change.
3. Finally, if the tag is not recognized and it is not already close, _then_ it can skip the closing tag.

As I mentioned in #464, I think the difficulty here is that htmlparser2 automatically tries to close these tags. It doesn't provide a signal (that I'm aware of) that the tag wasn't already closed.

abea (changes requested):
@potter7050 Thanks for the contribution. On a style note, I would suggest calling this option `autoCloseTags`. I would also ask for documentation of the new feature in the README file.

The bigger issue is that setting `useAutoCloseTag: false` does not only disabling closing unclosed tags. It also fails to close tags that are _already closed_ in the markup. See this test:

```javascript
    const wackyText = sanitizeHtml('<p>here\'s a string with a <wacky> tag.</p>', {
      useAutoClosingTag: false,
      disallowedTagsMode: 'escape'
    });
    assert.equal(wackyText, '<p>here\'s a string with a &lt;wacky&gt; tag.</p>');
```

This fails with output `<p>here's a string with a &lt;wacky&gt; tag.`.

Here we have a closed paragraph tag with an unclosed fake tag. The fake tag isn't closed, but the paragraph tag isn't closed either. If we were to call this `closeTags: false`, this would no longer be serving to sanitize HTML.

A fix for the original problem would need to 
1. Recognize that a tag is not recognized (disallowed). If recognized, there should be no change.
2. Identify if the unrecognized tag is already closed in the input. If it is already closed, there should be no change.
3. Finally, if the tag is not recognized and it is not already close, _then_ it can skip the closing tag.

As I mentioned in #464, I think the difficulty here is that htmlparser2 automatically tries to close these tags. It doesn't provide a signal (that I'm aware of) that the tag wasn't already closed.

PR #472, opened 2021-04-25
Support wildcards in allowedClasses (update to #157)

zhangbenber:
Support wildcards in `allowedClasses` like this:

```javascript
allowedClasses: {
  p: ["prefix-*"]
}
```

---

Useful especially for allowing the `language-xxx` syntax indicator classes in [highlight.js](https://highlightjs.org/).

Attempts have been made in #157, and here I tried to build up a glob map before the loop for better performance, the same as what was done for allowed Attributes. And the wildcard may match zero character for consistency with allowed Attributes too.

All tests passed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## 2.4.0 (2021-05-05):
+- Added support for class names with wildcards in `allowedClasses`.
+
 ## 2.3.3 (2021-03-19):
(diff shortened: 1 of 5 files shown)


boutell (changes requested):
Looks good — please add to the README and the changelog.

boutell (changes requested):
Please add to the changelog as well. Thanks!

zhangbenber:
> Please add to the changelog as well. Thanks!

Glad to, but version numbers and release date may change according to maintainers' schedule :)

abea:
True. We normally add new items under an `Unreleased` heading but that wasn't clear. We can take care of it.

boutell (approved):

boutell:
@abea would you mind taking care of the changelog and adding this to the release notes plan?

PR #474, opened 2021-05-01
Add sandboxed demos

george-thomas-hill:
I have edited README.md so that it links to ten sandboxed demos that I created using runkit.com (which is like JSFiddle, except for back-end code powered by Node.js).

I also fixed a few typos and one bug in sample code.

Full disclosure: I did this work as a requirement of a class taken at Oregon State University.

### DIFF
--- a/README.md
+++ b/README.md
@@ -346,6 +346,8 @@ const clean = sanitizeHtml(dirty, {
 });
 ```
 
+[Demo.](https://runkit.com/george-hill/608d48848ee2b1001aeaf005)
+
 The `simpleTransform` helper method has 3 parameters:
 
 ```js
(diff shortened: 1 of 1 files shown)


abea:
@george-thomas-hill Hi George. Sorry for not replying earlier. We really like and appreciate this contribution. Thanks for taking such an interesting in santize-html. We decided not to merge this in primarily because it is risky to rely on third-party URLs that are not maintained by the core team. They could expire or disappear one day and we might not realize it until much later.

george-thomas-hill:
@abea Thank you very much for the reply.

I understand.

Thank you for maintaining sanitize-html!

PR #496, opened 2021-09-21
Fix #486: Return an empty string when input is a nullish HTML

### LINKED ISSUE #486: Nullish HTML input should return an empty string.
## The problem to solve

Discussed in #102. Undefined or null input returns 'undefined' and 'null' strings.

## Proposed solution

We return an empty string if the input is nullish

## Alternatives

We could also throw an error in that case, making it the responsibility of the application using sanitize-html to ensure input isn't nullish. I'm not sure there's enough downside to the proposed solution to pass the buck, though.

postatum:
Note that nullish values also include 0, but according to [this comment](https://github.com/apostrophecms/sanitize-html/pull/102#issuecomment-872242667) of a repo member: 
> In any case it is up to the caller to pass a string.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## 2.5.2:
+- Nullish HTML input now returns an empty string. Nullish value may be explicit `null`, `undefined` or implicit `undefined` when value is not provided.
+
 ## 2.5.1 (2021-09-14):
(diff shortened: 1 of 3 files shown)


boutell:
If it has a `toString()` I think it should be left alone, `0` is an unusual input but no need to reject it. I suggest testing with `== null` instead. This matches only `null` and `undefined` which are the only cases that don't support `toString()` and thus have trouble.

postatum:
> If it has a toString() I think it should be left alone, 0 is an unusual input but no need to reject it. I suggest testing with == null instead. This matches only null and undefined which are the only cases that don't support toString() and thus have trouble.

I see. I've changed the test to `== null`.

boutell (approved):

boutell (changes requested):
Would you kindly add a changelog entry? Thank you!

postatum:
> Would you kindly add a changelog entry? Thank you!

Done. Please let me know if that's an OK description.

boutell (approved):
Thanks!

Issue #506, opened 2021-10-15
Trailing text is not preserved


IvanPizhenko:
## To Reproduce

1. Get sample project that demonstrates bug (2 files) from here: https://gist.github.com/IvanPizhenko/100c92f523a994fd425ff8f16420f12d
2. Run `npm install`
3. Run `node index.js`

## Expected behavior
I expect trailing text `Above` is preserved, i.e. sanitized output should be:
```
Below
<span>[image: image.png]</span>
Above
```

## Describe the bug
Trailing text `Above` is lost, i.e. I am getting:
```
Below
<span>[image: image.png]</span>
```

## Details

**Version of Node.js:** v12.22.6

**Server Operating System:** Linux

**Additional context:**
I am trying to upgrade `sanitize-html` from version `1.21.1` to latest `2.5.2` (at the moment of submitting this issue) and getting the above described issue.  According to my research (I have tried multiple versions) bug has appeared in version `1.27.3`.

alex-rantos:
Stumble upon the issue myself when upgraded to latest from `1.22`.

It's a bug on transformTags on `text` property which replaces content. (fyi: @tomholub)

IvanPizhenko:
@alex-rantos Do you know how to fix this?

alex-rantos:
Not really I just narrowed down the repro steps a bit further and shared it here.
However, I will try to give it a go once I find some free time but I cannot promise anything - not familiar with transformTags flow.

alex-rantos:
It's a regression from this commit https://github.com/alex-rantos/sanitize-html/commit/b218a722a3706de54b1ddcca0377e131102ec8fd

@abea the following condition fails to append the trailing text as `addedText = true` in this case.
``` js
else if (!addedText) {
  result += escaped;
}
```

IvanPizhenko:
Works correctly in the version 2.6.1. Thank you for the fix!

PR #521, opened 2021-12-04
Fix trailing text issue on transformTag 

alex-rantos:
Fixes issue #506

Fixes regression caused by https://github.com/apostrophecms/sanitize-html/pull/395 in which `addedText` was never reseted to false after the `ontext` function was called for that tag.

Solution: reset `addedText` boolean to false in the onclosetag function.

### DIFF
--- a/index.js
+++ b/index.js
@@ -582,6 +582,7 @@ function sanitizeHtml(html, options, _recursing) {
         result = tempResult + escapeHtml(result);
         tempResult = '';
       }
+      addedText = false;
     }
   }, options.parser);
   parser.write(html);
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -582,6 +582,7 @@ function sanitizeHtml(html, options, _recursing) {
         result = tempResult + escapeHtml(result);
         tempResult = '';
       }
+      addedText = false;
```
alex-rantos, line 585: addedText needs to reseted to false after each closing tag.

abea (changes requested):
This looks great. Thank you. Please add a CHANGELOG entry under an `## Unreleased` heading. Include a message like "Thanks to Alexandros Rantos for the contribution." (linking to your Github account if you'd like)

abea (approved):

PR #555, opened 2022-07-12
Fix protocol relative url in scripts tags #531

### LINKED ISSUE #531: Cannot use protocol-relative URL in script src attribute
It's not possible to add a protocol-relative URL in the src attribute from script tag.

At naughtyHref function it is handled, but after that this code is executed:

const parsed = new URL(value);

This code generates an exception when a protocol-relative URL is used, generating an exception making allowed value as false:
              
} catch (e) {
   allowed = false;
}

Finally, it makes the src attribute is removed from script tag.

This happens between lines 325 and 355.

paweljq:
## Summary

Allow relative url in script tags. Logic is the same like in iframe tag. 
* Closes  #531 - 

Proper test has been written to cover this situation. Test name: 

> Should allow protocol-relative URLs for script tag

*(Check at least one)*

- [x] Bug fix
- [ ] New feature
- [x] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

**Other information:**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Changelog
 
+- Protocol-relative URLs are properly supported for script tags
+
 ## 2.7.0 (2022-02-04)
 
 - Allows a more sensible set of default attributes on `<img />` tags. Thanks to [Zade Viggers](https://github.com/zadeviggers).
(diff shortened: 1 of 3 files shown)


boutell (approved):
Thanks.

PR #577, opened 2022-10-11
For allowedTags, stop treating falsy values other than false, same as false.

### LINKED ISSUE #176: {allowedTags:null} allows <script>
```js
'<script>alert(1)</script>' ==
sanitizeHtml(
    '<script>alert(1)</script>',
    { allowedTags: null });

'<script>alert(1)</script>' ==
sanitizeHtml(
    '<script>alert(1)</script>',
    { allowedTags: undefined });
```

The [docs](https://github.com/punkave/sanitize-html#what-if-i-want-to-allow-all-tags-or-all-attributes) say

> ### "What if I want to allow all tags or all attributes?"
>
> Simple! instead of leaving allowedTags or allowedAttributes out of the options, set either one or both to false:
>
> ```js
> allowedTags: false,
> allowedAttributes: false
> ```

The internal check checks whether `allowedTags` is falsey, not false.

Treating `null` equivalently to false is problematic since `null` is
much more likely as an output from a function that otherwise
returns an array than `false`, so treating `null` and `undefined`
as equivalent to `false` is a corner-case with very serious security consequences.

For example,

```js
const MY_POLICY = {
  allowedTags: computeAllowedTags()
};

function computeAllowedTags() {
  if (complexCondition) {
    return INLINE_ELEMENTS;
  } else if (anotherComplexCondition) {
    return BLOCK_AND_INLINE_ELEMENTS;
  } else if (adNauseam) {
    return FORMATTING_ELEMENTS_AND_IMAGES;
  }
  // NOTE: Missing return at bottom implies return of undefined
}
```

Since the behavior for `undefined` and `null`, `0`, `NaN`, `""` and other falsey values is not documented, I recommend either

* changing the code that fils in blanks:
  ```js
  options = extend(sanitizeHtml.defaults, options);
  ```
  to first remove any properties with falsey, but non-false values.
* and/or change the falsey checks
  ```js
  if (options.allowedTags && options.allowedTags.indexOf(name) === -1) {
  ```
  to check for `false`:
  ```js
  if (options.allowedTags !== false
      && (options.allowedTags || []).indexOf(name) === -1) {
  ```
  and similarly for `allowedAttributes`.

kedarchandrayan:
## Summary

Closes #176 

Run the following example snippet.
```
sanitizeHtml('<div><wiggly worms="ewww">hello</wiggly></div>', {
      allowedTags: undefined
    })
```

Before this change, we used to get the following output, ie. all tags allowed:
`'<div><wiggly worms="ewww">hello</wiggly></div>'`

After this change, we will get the following output, ie. no tag allowed:
`'hello'`

I have added new test cases to check the behaviour needed.

*(Check at least one)*

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

**Other information:**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
-- If allowedTags is falsey but not exactly `false`, then do not assume that all tags are allowed. Rather, allow no tags in this case, to be on a safer side. This fixes [issue #176](https://github.com/apostrophecms/sanitize-html/issues/176).
+- If allowedTags is falsy but not exactly `false`, then do not assume that all tags are allowed. Rather, allow no tags in this case, to be on a safer side. This fixes [issue #176](https://github.com/apostrophecms/sanitize-html/issues/176).
 
 ## 2.7.2 (2022-09-15)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+- If allowedTags is falsey but not exactly `false`, then do not assume that all tags are allowed. Rather, allow no tags in this case, to be on a safer side. This fixes [issue #176](https://github.com/apostrophecms/sanitize-html/issues/176).
+
 ## 2.7.2 (2022-09-15)
 
 - Closing tags must agree with opening tags. This fixes [issue #549](https://github.com/apostrophecms/sanitize-html/issues/549), in which closing tags not associated with any permitted opening tag could be passed through. No known exploit exists, but it's better not to permit this. Thanks to 
--- a/index.js
+++ b/index.js
@@ -117,7 +117,7 @@ function sanitizeHtml(html, options, _recursing) {
   // vulnerableTags
   vulnerableTags.forEach(function (tag) {
     if (
-      options.allowedTags && options.allowedTags.indexOf(tag) > -1 &&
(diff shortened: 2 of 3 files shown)


boutell:
This would be a bc break as it is not unreasonable to rely on the current behavior. Thus we'd have to force a major version change for it and I don't think it's really justified for this strictness. I would say that developers who have this concern should use the community-maintained TypeScript bindings to achieve the desired certainty.

kedarchandrayan:
Hello @boutell, we discussed the same on the issue and after your go-ahead, the PR was created.

Actually, this strictness is needed as due to some bug in the user's code, undefined can come as a value passed to the `allowedTags` parameter, as discussed in the issue.

Developers who are already using the community-maintained TypeScript bindings will have no effect. It will effect only those who have accidentally sent `null` or `undefined` or `0` in the value of `allowedTags`. These usages were wrong, so avoiding those should not be a problem as suggested by you in the issue discussion.

IMO, a major version update is not needed as it is not breaking anything which was promised in the readme.

Please refer the issue #176 for the earlier discussion.

Thanks,
Kedar Chandrayan

kedarchandrayan:
Thanks @boutell. Please let me know if you want any changes.

kedarchandrayan:
Hello @boutell, please let me know your comments and changes if needed.

boutell:
Thanks!

PR #596, opened 2022-12-17
feat(parseStyleAttributes): add option to skip style parsing [fix 547]

bertyhell:
This will fix https://github.com/apostrophecms/sanitize-html/issues/547

This PR introduces a new option:
```
options.parseStyleAttributes: boolean
```

By default set to true, to match the current behavior. But a user can set it to false, to skip parsing style tags.
This can avoid issues when the package is used in the browser.

```javascript
  it('Should ignore styles when options.parseStyleAttributes is false', function() {
    assert.equal(
      sanitizeHtml('<span style=\'color: blue; text-align: justify\'></span>', {
        allowedTags: false,
        allowedAttributes: {
          span: [ 'style' ]
        },
        allowedStyles: {
          span: {
            color: [ /blue/ ],
            'text-align': [ /left/ ]
          }
        },
        parseStyleAttributes: false
      }), '<span style="color: blue; text-align: justify"></span>'
    );
  });
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -637,11 +637,13 @@ This will transform `<div style="invalid-prop: non-existing-value">content</div>
 
 By default the parseStyleAttributes option is true.
 
-When you disable the parsing of the style attribute, the `allowedStyles` option is automatically ignored, and all styles will be allowed.
+When you disable parsing of the style attribute (`parseStyleAttributes: false`) and you pass in options for the allowedStyles property, an error will be thrown. This combination is not permitted.
 
 Note that by default this library also strips style tags when you run this package in the browser. Because of a postcss dependency. https://github.com/apostrophecms/sanitize-html/issues/547
 
-So we recommend to use this package on the server in a node environment. And if you really need to run it on the client in the browser, to disable parseStyleAttributes.
+we recommend sanitizing content server-side in a Node.js environment, as you cannot trust a browser to sanitize things anyway. Consider what a malicious user could do via the network panel, 
+the browser console, or just by writing scripts that submit content similar to what your JavaScript submits. But if you really need to run it on the client in the browser, 
+you may find you need to disable parseStyleAttributes. This is subject to change as it is [an upstream issue with postcss](https://github.com/postcss/postcss/issues/1727), not sanitize-html itself.
 
 ### Restricting deep nesting
 
--- a/README.md
+++ b/README.md
@@ -139,7 +139,8 @@ allowedSchemes: [ 'http', 'https', 'ftp', 'mailto', 'tel' ],
 allowedSchemesByTag: {},
 allowedSchemesAppliedToAttributes: [ 'href', 'src', 'cite' ],
 allowProtocolRelative: true,
-enforceHtmlBoundary: false
+enforceHtmlBoundary: false,
+parseStyleAttributes: true
 ```
 
 ### Common use cases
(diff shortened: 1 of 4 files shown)


boutell:
Thanks. Does `allowedStyles` still do anything when `parseStyleAttributes` is false? Is it misleading to suggest anything is done to sanitize styles in the presence of this setting?

bertyhell:
@boutell Good point, i added some extra text in the readme to make the user aware of this:

```
When you disable the parsing of the style attribute, the `allowedStyles` option is automatically ignored, and all styles will be allowed.
```

bertyhell:
@boutell fixed, when they are used together you get this error:
```
throw new Error('allowedStyles option cannot be used together with parseStyleAttributes: false.');
```

test/test.js
```diff
@@ -956,6 +956,35 @@ describe('sanitizeHtml', function() {
...
+        allowedAttributes: {
+          span: ['style']
+        },
+        allowedStyles: {
+          p: {
+            'text-align': [/^justify$/]
+          }
+        },
+        parseStyleAttributes: false
+      });
```
boutell, line 983: Need an "assert(false)" at the end of the "try" block so we confirm we didn't get past it.

boutell (changes requested):

bertyhell:
@boutell  fixed

README.md
```diff
@@ -624,6 +625,24 @@ This will transform `<disallowed>content</disallowed>` to `&lt;disallowed&gt;con
...
+
+```js
+parseStyleAttributes: false
+```
+
+This will transform `<div style="invalid-prop: non-existing-value">content</div>` to `<div style="invalid-prop: non-existing-value">content</div>` instead of stripping it: `<div>content</div>`
+
+By default the parseStyleAttributes option is true.
+
+When you disable the parsing of the style attribute, the `allowedStyles` option is automatically ignored, and all styles will be allowed.
```
boutell, line 640: This sentence and the next one are a little inaccurate now. If you set allowedStyles and parseStyleAttributes: false at the same time, an error is thrown — the combination is not permitted. That's what needs to be documented.

README.md
```diff
@@ -624,6 +625,24 @@ This will transform `<disallowed>content</disallowed>` to `&lt;disallowed&gt;con
...
+
+Note that by default this library also strips style tags when you run this package in the browser. Because of a postcss dependency. https://github.com/apostrophecms/sanitize-html/issues/547
```
boutell, line 642: I don't think this is really an accurate description of the problem, but I get that it is rather difficult to pin down and maybe doesn't belong in the README as it could get fixed upstream. So I suggest we lose this paragraph, but see my notes on the next one.

README.md
```diff
@@ -624,6 +625,24 @@ This will transform `<disallowed>content</disallowed>` to `&lt;disallowed&gt;con
...
+
+So we recommend to use this package on the server in a node environment. And if you really need to run it on the client in the browser, to disable parseStyleAttributes.
```
boutell, line 644: I would say "we recommend sanitizing content server-side in a Node.js environment, as you cannot trust a browser to sanitize things anyway. Consider what a malicious user could do via the network panel, the browser console, or just by writing scripts that submit content similar to what your JavaScript submits. But if you really need to run it on the client in the browser, you may find you need to disable `parseStyleAttributes`. This is subject to change as it is an upstream issue with `postcss`, not `sanitize-html` itself."

boutell (changes requested):
Thanks. We're down to README tweaks now...

bertyhell:
@boutell fixed

boutell (approved):

boutell:
Thanks!

PR #605, opened 2023-01-27
oops, we forgot to release parseStyleAttributes

boutell:
There was already a 2.8.1 release, probably the new publish attempt failed and I missed it because I was doing too many things at once.

No new functionality here, just releasing what we had already said was released in response to a ticket.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,12 @@
 # Changelog
 
+## 2.9.0 (2023-01-27)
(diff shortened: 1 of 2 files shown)


BoDonkey (approved):

boutell:
Actually we didn't forget, we just confused ourselves and others by not marking it "UNRELEASED" in the changelog. We'll do that next time.

PR #607, opened 2023-01-31
Fix auto-adding escaped closing tags

### LINKED ISSUE #464: Escaping input with unclosed, invalid tags returns it with unneeded escaped closing tags appended to end
## To Reproduce

From the Node REPL:
```
> const s = require('sanitize-html')
undefined
> s("here's a string with a <wacky> tag.", {disallowedTagsMode: "escape"})
"here's a string with a &lt;wacky&gt; tag.&lt;/wacky&gt;"
```

## Expected behavior

The offending tag is escaped, but no addition markup is created. For example, given the input `here's a string with a <wacky> tag`, I would expect the output to be: `here's a string with a &lt;wacky&gt; tag`

## Describe the bug

When sanitizing a string with `disallowedTagsMode: "escape"`, extra, unneeded (escaped) closing tags are appended to the end of the string.

## Details

**Version of Node.js:** 15.5.0
**Server Operating System:** Ubuntu 20.04

dliebner:
## Summary

Closes #464 
Adds `tagAllowed()` helper function.

Existing tests may need to be reviewed. I don't do a lot of PRs, sorry.

*(Check at least one)*

- [x] Bug fix
- [ ] New feature
- [x] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

- [ ] Related tests have been updated

**Other information:**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## UNRELEASED (2023-01-31)
+
+- Fix auto-adding escaped closing tags. In other words, do not add implied closing tags to disallowed tags when `disallowedTagMode` is set to any variant of `escape` -- just escape the disallowed tags that are present. This fixes [issue #464](https://github.com/apostrophecms/sanitize-html/issues/464). Thanks to [Daniel Liebner](https://github.com/dliebner)
+- Add `tagAllowed()` helper function which takes a tag name and checks it against `options.allowedTags` and returns `true` if the tag is allowed and `false` if it is not.
+
 ## 2.9.0 (2023-01-27)
 
 - Add option parseStyleAttributes to skip style parsing. This fixes [issue #547](https://github.com/apostrophecms/sanitize-html/issues/547). Thanks to [Bert Verhelst](https://github.com/bertyhell).
--- a/index.js
+++ b/index.js
@@ -566,17 +566,11 @@ function sanitizeHtml(html, options, _recursing) {
       frame.updateParentNodeMediaChildren();
       frame.updateParentNodeText();
 
-      if( name === 'wacky' ) {
-
-        console.log( tagAllowed(name), ['escape','recursiveEscape'].indexOf(options.disallowedTagsMode) >= 0 );
-
-      }
-
       if (
         // Already output />
         options.selfClosing.indexOf(name) !== -1 ||
         // Escaped tag
-        (tagAllowed(name) || ['escape','recursiveEscape'].indexOf(options.disallowedTagsMode) >= 0)
+        (!tagAllowed(name) && ['escape','recursiveEscape'].indexOf(options.disallowedTagsMode) >= 0)
       ) {
         if (skip) {
           result = tempResult;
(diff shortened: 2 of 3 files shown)


BoDonkey:
Hi @dliebner 
Thanks for giving this a go! As a first step, can you at least describe how your new function should work? Does the `tagAllowed()` take a boolean, or...? You can use the original example to provide the input and expected output. "<p>here is a string with a <wacky> tag.</p>"
Thanks!
Bob

dliebner:
`tagAllowed()` takes a tag name and checks it against `options.allowedTags` and returns `true` if the tag is allowed and `false` if it is not.

```
const wackyText = sanitizeHtml('<p>here\'s a string with a <wacky> tag.', {
	disallowedTagsMode: 'escape',
});
  
console.log( wackyText ); // expected: "<p>here's a string with a &lt;wacky&gt; tag.</p>"
```

BoDonkey:
Hi @dliebner,
I'm still not quite getting it. I wrote the following test:
```
it('Should allow unclosed tags', function() {
    assert.equal(
      sanitizeHtml('<p>Text with a <wacky> tag.</p>', {
        allowedTags: [ 'wacky' ],
        disallowedTagsMode: 'escape'
      }), '<p>Text with a <wacky> tag.</p>'
    );
  });
```
This test fails with the output being: `&lt;p&gt;Text with a <wacky> tag.</wacky>&lt;/p&gt;`
While I expected: `&lt;p&gt;Text with a &lt;wacky&gt; tag.&lt;/p&gt;` based on the original issue.
Any additional feedback?
Thanks,
Bob

dliebner:
In your test, you set your `allowedTags` to `['wacky']`, meaning only the `wacky` tag is allowed. All other tags will be escaped using HTML entities.  The opening `<p>` is correctly sanitized as it is not allowed. The opening `<wacky>` tag is correctly preserved as it is allowed, and it is correctly auto-closed as it is allowed. The closing `</p>` tag is correctly sanitized because it is not allowed (and is literally included in the input, as opposed to being only implied by the presence of an opening tag).

BoDonkey (changes requested):
Now I understand the functionality. Thanks! Can you update your PR with a test in the `/test/test.js` file? It can essentially be the test that I wrote, but please add a better descriptive sentence. Can you also update the `CHANGELOG.md` file, please? At the top add a section called 'UNRELEASED' and then a description of your change.

BoDonkey (changes requested):
Thanks for making the changes. We are almost there! 
Two items:
1) If you give the command `npm run test`, it will fail due to linting code errors. Please fix the ones you introduced.
2) While your tests are passing, if the first or second ones were to fail, we wouldn't have proper notification since they are combined together in a single `it()`. Please separate these into two test statements.
Sorry to nitpick and thanks for doing this.

BoDonkey (approved):
Thanks for the changes! It looks good, but today is a release day, so I won't merge it until tomorrow. The changes will be published on our next release day, which is Feb. 15.

PR #624, opened 2023-06-05
feat: remove empty non-boolean attributes

### LINKED ISSUE #123: Empty attributes
Hi, I notice after doing a transformation the empty attributes are not valid any more Example of the src element in an image tag
**Before**
``
**After**
``

Not sure if it is a bug, but can you help me?

dylanarmstrong:
## Summary

This causes sanitize-html to remove empty attributes that are known to be non-boolean. It provides an exhaustive list of all known attributes taken from: https://html.spec.whatwg.org/multipage/indices.html#attributes-3

* User controlled via `options.nonBooleanAttributes`
* Enabled by default, and can be disabled by passing in `nonBooleanAttributes: []`
* Can handle all empty attributes via `nonBooleanAttributes: ['*']`

Closes #123 

Test with empty attributes such as:

```
<input checked type="checkbox" />
```

Will not remove anything.

```
<input checked form type="checkbox" />
```

Will transform into the first example.

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

### DIFF
--- a/README.md
+++ b/README.md
@@ -212,7 +212,7 @@ allowedTags: false,
 allowedAttributes: false
 ```
 
-#### "What if I want to allow empty attributes?"
+#### "What if I want to allow empty attributes, even for cases like href that normally don't make sense?"
 
 Very simple! Set `nonBooleanAttributes` to `[]`.
 
@@ -220,9 +220,13 @@ Very simple! Set `nonBooleanAttributes` to `[]`.
 nonBooleanAttributes: []
 ```
 
-#### "What if I want to delete all empty attributes?"
+#### "What if I want to remove all empty attributes, including valid ones?"
 
 Also very simple! Set `nonBooleanAttributes` to `['*']`.
+
+**Note**: This will break common valid cases like `checked` and `selected`, so this is
+unlikely to be what you want. For most ordinary HTML use, it is best to avoid making
+this change.
 
 ```js
 nonBooleanAttributes: ['*']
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -167,6 +212,22 @@ allowedTags: false,
 allowedAttributes: false
 ```
 
+#### "What if I want to allow empty attributes?"
```
boutell, line 215: This implies they are not allowed by default. They are allowed, just only where it makes sense.

This could be a better heading:

"What if I want to allow empty attributes, even for cases like `href` that normally don't make sense?"

README.md
```diff
@@ -167,6 +212,22 @@ allowedTags: false,
...
+nonBooleanAttributes: []
+```
+
+#### "What if I want to delete all empty attributes?"
+
+Also very simple! Set `nonBooleanAttributes` to `['*']`.
+
+```js
+nonBooleanAttributes: ['*']
+```
```
boutell, line 229: Add a note that this will break common valid cases like `checked` and `selected` so for most ordinary HTML use cases it is best to avoid making this change.

boutell (changes requested):
Great stuff. Minor documentation requests 🙏

README.md
dylanarmstrong: Updated

README.md
dylanarmstrong: Updated

dylanarmstrong:
Thanks for feedback, updated documentation.

boutell (approved):
Good stuff, thank you!

PR #646, opened 2024-02-04
allow svg element

gkumar9891:
In this PR I have added a test case for the SVG element to prevent the element from lowercased and same for the attributes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,11 +2,9 @@
 
 ## UNRELEASED
 
-- Clarified the use of SVGs with a new test and changes to documentation. 
-
-## UNRELEASED
-
 - Introduced the `allowedEmptyAttributes` option, enabling explicit specification of empty string values for select attributes, with the default attribute set to `alt`.
+
+- Clarified the use of SVGs with a new test and changes to documentation. 
 
 ## 2.11.0 (2023-06-21)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## UNRELEASED
 
+- Clarified the use of SVGs with a new test and changes to documentation. 
+
+## UNRELEASED
+
 - Introduced the `allowedEmptyAttributes` option, enabling explicit specification of empty string values for select attributes, with the default attribute set to `alt`.
 
 ## 2.11.0 (2023-06-21)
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## UNRELEASED
 
+- Clarified the use of SVGs with a new test and changes to documentation.
```
BoDonkey, line 5: This can be moved under the other unreleased heading. We will change the 'UNRELEASED' to a version number when we publish all the changes.

README.md
```diff
@@ -265,6 +265,19 @@ allowedAttributes: {
 
 With `multiple: true`, several allowed values may appear in the same attribute, separated by spaces. Otherwise the attribute must exactly match one and only one of the allowed values.
 
+#### "What if I want to prevent attributes and tags from being truly lowercase?"
```
BoDonkey, line 268:
```suggestion
#### "What if I want to maintain the original case for SVG elements and attributes?"
```

README.md
```diff
@@ -265,6 +265,19 @@ allowedAttributes: {
...
+
+if you got any tag like `linearGradient` and you want to prevent that tag from being lowercased so you have to pass `lowerCaseTags: false` in `parser` option and same for if you want to prevent any attribute like `viewBox` in svg tag from being lowercased so you have to pass `lowerCaseAttributeNames: false` in parser option.
```
BoDonkey, line 270:
```suggestion
If you're incorporating SVG elements like `linearGradient` into your content and notice that they're not rendering as expected due to case sensitivity issues, it's essential to prevent `sanitize-html` from converting element and attribute names to lowercase. This situation often arises when SVGs fail to display correctly because their case-sensitive tags, such as `linearGradient` and attributes like `viewBox`, are inadvertently lowercased.

To address this, ensure you set `lowerCaseTags: false` and `lowerCaseAttributeNames: false` in the parser options of your sanitize-html configuration. This adjustment stops the library from altering the case of your tags and attributes, preserving the integrity of your SVG content.
```

BoDonkey (changes requested):
I suggested a rewrite of your README section to better capture the use case as suggested by the original issue. Feel free to change it if you find any details left out.

BoDonkey:
Thanks for doing this! Small comments.

README.md
gkumar9891: I did the change

README.md
gkumar9891: I did the changes.

BoDonkey (approved):

PR #647, opened 2024-02-06
added completelyDiscard option

gkumar9891:
Added 'completelyDiscard: true' option to remove the content in html.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,9 @@
 
 - Introduced the `allowedEmptyAttributes` option, enabling explicit specification of empty string values for select attributes, with the default attribute set to `alt`.
 
-- Clarified the use of SVGs with a new test and changes to documentation. 
+- Clarified the use of SVGs with a new test and changes to documentation.
+
+- Added completelyDiscard option to remove the content also in html
 
 ## 2.11.0 (2023-06-21)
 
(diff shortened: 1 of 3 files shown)


BoDonkey (changes requested):
Thanks for tackling these issues!

Your test coverage is only for a simple sanitization case. When testing a more complex case, such as when the tag inside is a nonTextTag, your modified code fails.
For example:
```
it('should completely remove disallowed tags with nested content', () => {
  const inputHtml = '<div><p>Allowed content</p><script>var x = "Disallowed script";</script><span>More allowed content</span></div>';
  const expectedOutput = '<div><p>Allowed content</p><span>More allowed content</span></div>';
  
  const sanitizedHtml = sanitizeHtml(inputHtml, {
    allowedTags: ['div', 'p', 'span'], // 'script' tag is disallowed
    disallowedTagsMode: 'discard',
    completelyDiscard: true
  });
  
  assert.equal(sanitizedHtml, expectedOutput);
});
```
While this shows a test for nested content, I think it is failing because of the use of the `script` tag. If you change that to an `h2` I think it will pass. I'm also wondering what happens if a user passes `completelyDiscard: false`.

Basically, I think you need to write some tests for more edge cases like nesting, option = false, self-closing tags, and make sure your code passes.

Issue #49, opened 2015-04-02
Add a space where two tags met


danschumann:
I'm wondering if there is a way to create a space wherever tags were.

Picture this:

`<div>Some sentence.</div><div>Some other Sentence</div>`

It converts to `Some sentence.Some other Sentence` when I run

`text = sanitizeHtml(text, {allowedTags:[], allowedAttribute: {}});`

Is there an option to add whitespace so the output is better: `Some sentence. Some other Sentence`

dgrad:
I think to do this properly you'll need a list of block tags (or a list of inline tags). You want to add a space wherever a block tag ends (actually it should probably be a newline character and let the browser convert it to space), but not where an inline tag ends (e.g. `<div>foo</div><div>bar</div>` should convert to `foo bar`, but `<span>foo</span><span>bar</span>` should convert to `foobar`).

boutell:
I agree, and there should be an option to override that list.

I'd take a pull request for this one.

On Fri, Nov 27, 2015 at 12:42 PM, Daniel Grad <notifications@github.com>
wrote:

> I think to do this properly you'll need a list of block tags (or a list of
> inline tags). You want to add a space wherever a block tag ends (actually
> it should probably be a newline character and let the browser convert it to
> space), but not where an inline tag ends (e.g.
> <div>foo</div><div>bar</div> should convert to foo bar, but
> <span>foo</span><span>bar</span> should convert to foobar).
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/punkave/sanitize-html/issues/49#issuecomment-160181276>
> .
>



-- 


*THOMAS BOUTELL, *DEV & OPS
P'UNK AVENUE | (215) 755-1330  |  punkave.com

SystemDisc:
Until this gets implemented, a messy hack would be:
```javascript
text = text.replace(/>/g, '> ');
text = sanitizeHtml(text, {allowedTags:[]});
```

r-custodio:
up!!

Is this still on?

boutell:
@r-custodio As mentioned, I'd take a PR for this. Unfortunately as maintainer I can't necessarily implement every feature.

greghub:
@abea this is labeled as `seeking contributions` but closed. Is it still something you'd accept a PR for?

abea:
[...]

PR #692, opened 2025-02-02
feat: add tag open, tag close, and text events

### LINKED ISSUE #691: Allow hooking into onopentag and onclosetag
## The problem to solve

I'm trying to solve the issue described here https://github.com/apostrophecms/sanitize-html/issues/49. I think I'm quite close to doing it correctly, but current `sanitize-html` functions and configuration seem too limiting to express this.

Imagine performing the following two modifications to HTML:
1. insert a space before the first text block after a disallowed element has just opened
2. insert a space before the first text block after a disallowed element has just closed

For example (assume we disallow all `<div>`):
```html
<div>hello</div>there -> "hello there"
hello<div>there</div> -> "hello there"
<div>hello</div><div>there</div> -> " hello there"
<div><div>hello</div>there</div> -> " hello there"
```

Especially note scenarios like this (where we can't simply add a space before each closing `</div>`):
```html
<div><text>this is a text block</text><text>this is a text block</text></div>
```

I don't think this is currently possible (but correct me if I'm wrong).

## Proposed solution

Allow hooking into `ontagopen` and `ontagclose` events and just pass `tag` names into the listeners like this:
```js
sanitizeHtml(
  '<p>This is <a href="http://www.linux.org"></a><br/>Linux</p>',
  {
    onOpenTag: function(tag) {
      console.log(`${tag} opened`);
    },
    onCloseTag: function(tag) {
      console.log(`${tag} closed`);
    },
  }
);
```

## Alternatives

Tried all sorts of hacky ways of using `transformTags` and `exclusiveFilter` and `textFilter` but in combination with `discard` it doesn't work.

Maybe an alternative could be if `exclusiveFilter` could allow applying the `discard` mode, then it *may* be possible to hook into the `onclosetag` indirectly. But that's hacky.

naktinis:
## Summary

Allows hooking into tag tag open, tag close, and text events of the parser to enable more advanced transformations such as inserting a space between two removed block-level tags with preserved text.

Closes #691.

Added a simple test in the test suite, but you can also test manually by passing the `onOpenTag`, `onCloseTag`, and `onText` listeners in the sanitizer options.

*(Check at least one)*

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

### DIFF
--- a/README.md
+++ b/README.md
@@ -798,7 +798,7 @@ const sanitizedHtml = sanitizeHtml(
 );
 ```
 
-In this example we are tracking when a removed tag has opened or closed and modify text that follows it. The example should produce:
+In this example, we are setting a flag when a tag that will be removed has been opened or closed. Then we use the `textFilter` to modify the text to include spaces. The example should produce:
 ```
 Text's here it's here it's there and <b>also</b> here.
 ```
--- a/README.md
+++ b/README.md
@@ -786,11 +786,11 @@ const sanitizedHtml = sanitizeHtml(
   'There should be<div><p>spaces</p></div>between <b>these</b> words.',
   {
     allowedTags,
-    onOpenTag: (tag) => {
-      addSpace = !allowedTags.includes(tag);
+    onOpenTag: (tagName, attribs) => {
+      addSpace = !allowedTags.includes(tagName);
     },
-    onCloseTag: (tag) => {
-      addSpace = !allowedTags.includes(tag);
+    onCloseTag: (tagName, isImplied) => {
+      addSpace = !allowedTags.includes(tagName);
     },
     textFilter: (text) => {
       if (addSpace) {
--- a/README.md
+++ b/README.md
@@ -778,7 +778,7 @@ For example, you may want to add spaces around a removed tag, like this:
 const allowedTags = [ 'b' ];
 let addSpace = false;
 const sanitizedHtml = sanitizeHtml(
-  'Text&#39;s here<div>it&#39;s here</div><div><p>it&#39;s there</p></div>and <b>also</b> here.',
+  'There should be<div><p>spaces</p></div>between <b>these</b> words.',
   {
     allowedTags,
     onOpenTag: (tag) => {
@@ -800,7 +800,7 @@ const sanitizedHtml = sanitizeHtml(
 
 In this example, we are setting a flag when a tag that will be removed has been opened or closed. Then we use the `textFilter` to modify the text to include spaces. The example should produce:
 ```
-Text's here it's here it's there and <b>also</b> here.
+There should be spaces between these words.
 ```
 
 ## About ApostropheCMS
--- a/README.md
+++ b/README.md
@@ -772,11 +772,11 @@ For more advanced filtering you can hook directly into the parsing process using
 
 The `onOpenTag` event is triggered when an opening tag is encountered. It has two arguments:
 - `tagName`: The name of the tag.
-- `attribs`: An object containing the tag's attributes, i.e. { src: "/path/to/tux.png" }.
+- `attribs`: An object containing the tag's attributes, e.g. `{ src: "/path/to/tux.png" }`.
 
 The `onCloseTag` event is triggered when a closing tag is encountered. It has the following arguments:
 - `tagName`: The name of the tag.
-- `isImplied`: A boolean indicating whether the closing tag is implied (i.e. `<p>foo<p>bar`) or explicit (i.e. `<p>foo</p><p>bar</p>`).
+- `isImplied`: A boolean indicating whether the closing tag is implied (e.g. `<p>foo<p>bar`) or explicit (e.g. `<p>foo</p><p>bar</p>`).
 
 For example, you may want to add spaces around a removed tag, like this:
 ```js
--- a/README.md
+++ b/README.md
@@ -805,7 +805,7 @@ const sanitizedHtml = sanitizeHtml(
 
 In this example, we are setting a flag when a tag that will be removed has been opened or closed. Then we use the `textFilter` to modify the text to include spaces. The example should produce:
 ```
-There should be spaces between these words.
+There should be spaces between <b>these</b> words.
 ```
 
 ## About ApostropheCMS
--- a/README.md
+++ b/README.md
@@ -768,10 +768,15 @@ This will prevent the user from nesting tags more than 6 levels deep. Tags deepe
 
 ### Advanced filtering
 
-For more advanced filtering you can hook directly into the parsing process using three events:
-- `onOpenTag(name, attribs)`: Triggered when an opening tag is encountered.
-- `onText(text)`: Triggered when a text node is encountered.
-- `onCloseTag(name, isImplied)`: Triggered when a closing tag is encountered.
+For more advanced filtering you can hook directly into the parsing process using tag open and tag close events.
+
+The `onOpenTag` event is triggered when an opening tag is encountered. It has two arguments:
+- `tagName`: The name of the tag.
+- `attribs`: An object containing the tag's attributes, i.e. { src: "/path/to/tux.png" }.
+
+The `onCloseTag` event is triggered when a closing tag is encountered. It has the following arguments:
+- `tagName`: The name of the tag.
+- `isImplied`: A boolean indicating whether the closing tag is implied (i.e. `<p>foo<p>bar`) or explicit (i.e. `<p>foo</p><p>bar</p>`).
 
 For example, you may want to add spaces around a removed tag, like this:
 ```js
--- a/README.md
+++ b/README.md
@@ -766,6 +766,43 @@ nestingLimit: 6
 
 This will prevent the user from nesting tags more than 6 levels deep. Tags deeper than that are stripped out exactly as if they were disallowed. Note that this means text is preserved in the usual ways where appropriate.
 
+### Advanced filtering
+
+For more advanced filtering you can hook directly into the parsing process using three events:
+- `onOpenTag(name, attribs)`: Triggered when an opening tag is encountered.
+- `onText(text)`: Triggered when a text node is encountered.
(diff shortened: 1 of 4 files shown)


BoDonkey:
I haven't tried out the code changes, yet, or thought very deeply about the new test. However, can you please add to the CHANGELOG and also add some text to the README so that a user would know how to use the new options you have added? Thanks!

BlueManCZ:
This would be useful if you were actually using the return values of these functions. Right now, you're just calling them and ignoring the potential results. This has no effect on the result transformation. Am I missing something? Also, it would be nice to add some unit tests with real-world use cases, not just a call counter. (maybe #49 as you mention in the relevant issue #691).

naktinis:
I would use it to track the parsing state outside of the callbacks.

First of all, here's what would happen if I didn't apply any transformations:
```js
sanitizeHtml(
  '<div>I&#39;m here</div><div>I&#39;m there</div>',
  {
    allowedTags: [],
  },
);
// Returns "I'm hereI'm there"
```

Here's what happens if I insert a space before each text element:
```js
sanitizeHtml(
  '<div>I&#39;m here</div><div>I&#39;m there</div>',
  {
    allowedTags: [],
    textFilter: (text) => {
      return ' ' + text;
    },
  },
);
// Returns "I ' m here I ' m there"
```

Finally, a solution using the proposed `onOpenTag` and `onCloseTag`:
```js
let addSpace = false;
sanitizeHtml(
  '<div>I&#39;m here</div><div>I&#39;m there</div>',
  {
    allowedTags: [],
    onOpenTag: (tag) => {
      addSpace = true;
    },
    onCloseTag: (tag) => {
      addSpace = true;
    },
    textFilter: (text) => {
      if (addSpace) {
        addSpace = false;
        return ' ' + text;
      }
      return text;
    },
  },
);
// Returns "I'm here I'm there"
```

So this is the only way I could find to produce this end result correctly (of course, after introducing the new events). I tried hooking into `transformTags` and `exclusiveFilter` to detect tag opening and closing, but the current logic prevented it from fully working because they don't really correspond exactly to tag open and tag close events (plus even if it was possible, it still feels like a convoulted and unintuitive way of tracking this state).

BoDonkey:
Sorry I let this drop. @BlueManCZ is right about the tests needing update. Just calling those counters isn't that informative. You should show that your changes both address the original issue and don't break any other functionality.

naktinis:
Added a new test for a real-world use case as described above. Anything else you'd like to be tested?

BoDonkey:
I'll check a little later today.

BoDonkey:
I'm really sorry for letting this fall through the cracks again. For your tests, things we really want and need to see are:
1) Does it solve the original issue `<div>Some sentence.</div><div>Some other Sentence</div>` => `Some sentence. Some other Sentence`
2) Do each of your hooks get the expected content delivered to the callback -> so tag name, attributes, and/or text?
I would also like to see a test where some tags are stripped and others are not producing the expected output.

Additionally, we are going to need a bit of documentation to let the users know how the parse order might impact output and what parameters are being passed to the callbacks. I can help out with this last bit.

naktinis:
@BoDonkey no worries, thank you for looking into this.

1. I actually already added a test for that in the previous commit (the test is called `should insert spaces between removed tags whose content we keep`), but let me know if that's not what you expected.

2. I extended the call test to check the passed arguments. Also, extended the space insertion test to test the case where some tags are allowed while others are removed, and also tag nesting.

In terms of documentation, I imagine this could be in a section called "Parsing events" or "Advanced filtering" and talk about tracking parsing state outside of the core sanitizer logic. It could also use this space insertion logic as a practical example.  Let me know your thoughts overall and if you'd like me to propose text for it.

BoDonkey:
@naktinis A section called "Advanced filtering" seems good. When you add a code example, can you use a simpler example than was used with your test? Perhaps the original issue example. I'm afraid that people will become confused by the HTML entities.

naktinis:
@BoDonkey submitted a proposal.

README.md
```diff
@@ -766,6 +766,43 @@ nestingLimit: 6
...
+- `onOpenTag(name, attribs)`: Triggered when an opening tag is encountered.
+- `onText(text)`: Triggered when a text node is encountered.
+- `onCloseTag(name, isImplied)`: Triggered when a closing tag is encountered.
+
+For example, you may want to add spaces around a removed tag, like this:
+```js
+const allowedTags = [ 'b' ];
+let addSpace = false;
+const sanitizedHtml = sanitizeHtml(
+  'Text&#39;s here<div>it&#39;s here</div><div><p>it&#39;s there</p></div>and <b>also</b> here.',
```
BoDonkey, line 781: I would prefer not to have HTML entities in this example. It makes it harder to read and understand.

README.md
```diff
@@ -766,6 +766,43 @@ nestingLimit: 6
...
+      }
+      return text;
+    }
+  }
+);
+```
+
+In this example we are tracking when a removed tag has opened or closed and modify text that follows it. The example should produce:
+```
+Text's here it's here it's there and <b>also</b> here.
```
BoDonkey, line 803: Same here with regards to the `'` and clarity.

README.md
naktinis: Got it. Will simplify. I added them deliberately because when they are present `htmlparser2` seems to produce two text nodes in a row, and we need to make sure no extra space is inserted before the second text node. But maybe for this simple example no need to overthink it.

README.md
BoDonkey, line 1: For the documentation, can you please add in a description of what the parameters are for each new function? Also, I'm struggling a bit with a use case for the `onText(text)`. I can see having a complete set of hooks "feels" reasonable, but in practice I'm not sure when you would use `onText()`. If you can think of one, we should document it so that users know when/how to use it.

naktinis: Removed `onText` because it's not necessary for my use case at this time. Also documented the attributes. Let me know if there's anything else you'd like to update.

README.md
```diff
@@ -766,6 +766,48 @@ nestingLimit: 6
...
+
+For example, you may want to add spaces around a removed tag, like this:
+```js
+const allowedTags = [ 'b' ];
+let addSpace = false;
+const sanitizedHtml = sanitizeHtml(
+  'There should be<div><p>spaces</p></div>between <b>these</b> words.',
+  {
+    allowedTags,
+    onOpenTag: (tag) => {
```
BoDonkey, line 789: I'm debating whether it might be clearer in your example if you were to use `tagName` instead of `tag` and if we should also show the second argument - although it isn't used. What do you think?

test/test.js
```diff
@@ -1720,4 +1720,46 @@ describe('sanitizeHtml', function() {
 
     assert.equal(sanitizedHtml, expectedOutput);
   });
+  it('should call onOpenTag and onCloseTag callbacks', () => {
```
BoDonkey, line 1723: I'm wondering if we should modify this test to include a tag with an implied close?

README.md
naktinis: Updated

test/test.js
naktinis: Included an implicitly closed tag

BoDonkey:
Looks pretty good. I'm going to run this past our CTO, but (I know I'm a pain in the butt!) can you also update the changelog? Just add an `UNRELEASED` at the top and your changelog message. Thanks and sorry!

naktinis:
No worries. Updated the changelog.

README.md
```diff
@@ -781,6 +781,48 @@ nestingLimit: 6
...
+- `attribs`: An object containing the tag's attributes, e.g. `{ src: "/path/to/tux.png" }`.
+
+The `onCloseTag` event is triggered when a closing tag is encountered. It has the following arguments:
+- `tagName`: The name of the tag.
+- `isImplied`: A boolean indicating whether the closing tag is implied (e.g. `<p>foo<p>bar`) or explicit (e.g. `<p>foo</p><p>bar</p>`).
+
+For example, you may want to add spaces around a removed tag, like this:
+```js
+const allowedTags = [ 'b' ];
+let addSpace = false;
```
boutell, line 799: This example might struggle with nested tags in some situations, because `addSpace` will get reset and then it doesn't get "popped" back to its previous value when we return to the enclosing tag's context. I think that could be overcome in the example without changing your API, but that depends on how the API will handle tags that do not need or have a closing tag like `link`, do they fire only `onOpenTag`? If so it would be very difficult to get this 100% right. What are your thoughts on this?

boutell (changes requested):

README.md
naktinis: Could you give an example HTML snippet you have in mind? Would be easier for me to reason about it.

Regarding the example, this is not necessarily meant to be 100% production-ready and cover all possible edge cases, but rather to illustrate new possibilities that open up with these event handlers. In a real-life scenario people may want to have a special list of tags or custom rules where this applies (e.g. only block-level tags, only visible tags, a subset of these etc.)

boutell:
I do hear that about the simple example. My concern is that it be able to correctly figure out where we are in the tag stacking context in all cases, as many uses of this API would need that, even if it's taking me a minute to come up with a scenario where your use case needs it (I do think it's possible)

Something like:

```
<p>
  <img src="/something.jpg">
</p>
```

How many tag-opens and tag-closes does that fire?

naktinis:
@boutell I just checked and in this cases the fired events and their parameters will be:
```
onOpenTag "p" {}
onOpenTag "img" { src: "/something.jpg" }
onCloseTag "img" true
onCloseTag "p" false
```

boutell (changes requested):
OK, great!

In that case, I would just suggest including a quick note that the example is not meant to be production-ready and for that for some use cases you might need to keep a stack of currently open tags, using the open and close events to push and pop items on the stack.

Other than that I'm good to go. Appreciate the contribution!

naktinis:
@boutell I added a README section about the example not being production-ready.

boutell (approved):
Thank you!

PR #695, opened 2025-03-03
feat: exclusive filter discard tag only

### LINKED ISSUE #694: Anchor tags without `href` should not be kept
Considering these options:

```js
{
    allowedTags: ["p", "strong", "em", "code", "s", "a"],
    allowedAttributes: {
        a: ["href"],
    },
    allowedSchemes: ["http", "https"],
}
```

The following HTML:

```html
<a href="javascript:alert('evil')">good site</a>
```

gets transformed into this:

```html
<a>good site</a>
```

instead of just

```html
good site
```

or being removed completely.

rChaoz:
## Summary

Add a method to keep inner content when discarding tags with exclusive filter (can return `"excludeTag"` instead of `true`). Closes #694

This PR also includes a couple of _tiny_ refactorings I saw while looking through the code, go commit by commit to see the changes separated.

See the example from the readme - run the following code:

```js
sanitizeHtml(
  'This is <a href="javascript:alert(123)">bad link</a> and a <a href="https://www.linux.org">good link</a>',
  {
    exclusiveFilter: function(frame) {
      // the href attribute is removed by the URL protocol check
      return frame.tag === 'a' && !frame.attribs.href ? 'excludeTag' : false;
    }
  }
)
```

In the current version, this outputs

> This is a  and a <a href="https://www.linux.org">good link</a>

With this PR, it outputs

> This is a bad link and a <a href="https://www.linux.org">good link</a>

*(Check at least one)*

- [ ] Bug fix
- [x] New feature
- [x] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -397,7 +397,7 @@ describe('sanitizeHtml', function() {
     );
   });
 
-  it('should keep inner text when exclusiveFilter returns "tag"', function() {
+  it('should keep inner text when exclusiveFilter returns "excludeTag"', function() {
     assert.strictEqual(
       sanitizeHtml('<p>These links <a href="javascript:alert(123)">hack</a> <a href="data:image/gif;base64,R0lGODlhAQABAAAAACH5BAEKAAEALAAAAAABAAEAAAICTAEAOw==">more hack</a> have disallowed href protocols</p>', {
         exclusiveFilter: function (frame) {
@@ -408,7 +408,7 @@ describe('sanitizeHtml', function() {
     );
   });
 
-  it('should keep inner tags when exclusiveFilter returns "tag"', function() {
+  it('should keep inner tags when exclusiveFilter returns "excludeTag"', function() {
     assert.strictEqual(
       sanitizeHtml('This div is bad <div class="bad">but its <strong>content</strong><p>should be kept <em>as-is</em></p></div>.', {
         exclusiveFilter: function (frame) {
@@ -419,7 +419,7 @@ describe('sanitizeHtml', function() {
     );
   });
 
-  it('should work with escaped tags when exclusiveFilter returns "tag"', function () {
+  it('should work with escaped tags when exclusiveFilter returns "excludeTag"', function () {
     assert.strictEqual(
       sanitizeHtml('<strong>hello</strong> <wiggly>there. <em>General Kenobi</em></wiggly>!', {
         disallowedTagsMode: 'escape',


boutell (changes requested):
I think this is great! However, I think we should be more explicit and check for a return value of `excludeTag`. This is to avoid any confusion over whether we're *keeping* just the tag, or *excluding* just the tag. With that small change I'm on board. Thank you for the contribution!

boutell (changes requested):
Thanks for the fast turnaround! But I think you missed it in the text part of your README updates.

rChaoz:
Also created a types PR here: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/72103
Yes, I updated the PR example but not the next, mb

test/test.js
```diff
@@ -384,6 +384,53 @@ describe('sanitizeHtml', function() {
...
+      sanitizeHtml('<p>These links <a href="javascript:alert(123)">hack</a> <a href="data:image/gif;base64,R0lGODlhAQABAAAAACH5BAEKAAEALAAAAAABAAEAAAICTAEAOw==">more hack</a> have disallowed href protocols</p>', {
+        exclusiveFilter: function (frame) {
+          return frame.tag === 'a' && !frame.attribs.src ? 'excludeTag' : false;
+        }
+      }),
+      '<p>These links hack more hack have disallowed href protocols</p>'
+    );
+  });
+
+  it('should keep inner tags when exclusiveFilter returns "tag"', function() {
```
boutell, line 411: returns "excludeTag"

test/test.js
```diff
@@ -384,6 +384,53 @@ describe('sanitizeHtml', function() {
...
+      sanitizeHtml('This div is bad <div class="bad">but its <strong>content</strong><p>should be kept <em>as-is</em></p></div>.', {
+        exclusiveFilter: function (frame) {
+          return frame.tag === 'div' && frame.attribs.class && /\bbad\b/.test(frame.attribs.class) ? 'excludeTag' : false;
+        }
+      }),
+      'This div is bad but its <strong>content</strong><p>should be kept <em>as-is</em></p>.'
+    );
+  });
+
+  it('should work with escaped tags when exclusiveFilter returns "tag"', function () {
```
boutell, line 422: returns "excludeTag"

test/test.js
```diff
@@ -384,6 +384,53 @@ describe('sanitizeHtml', function() {
...
+        {
+          exclusiveFilter: function () {
+            throw Error('this should not run');
+          }
+        }),
+      'this tag is discarded'
+    );
+  });
+
+  it('should keep inner text when exclusiveFilter returns "tag"', function() {
```
boutell, line 400: returns "excludeTag"

boutell (changes requested):

BoDonkey:
@rChaoz This is looking good. I'm not sure if this is ready for re-review and whether you are allowed to hit the request review button at the top?

rChaoz:
It was ready, I forgot to click it, my bad

boutell (approved):
Thank you for contributing!

Issue #717, opened 2025-11-20
Library automatically transforms escaped quotes to quotes


a2937:
PLEASE NOTE: make sure the bug exists in the latest patch level of the project. For instance, if you are running a 2.x version of Apostrophe, you should use the latest in that major version to confirm the bug.

## To Reproduce

Step by step instructions to reproduce the behavior:
Run the following code and check the output. 

```js
  const inputHtml = '<code>&apos;</code>';
  const message = sanitizeHtml(inputHtml, {
    allowedTags: [ 'b', 'i', 'em', 'strong', 'code', 'wbr' ]
  });
```

## Expected behavior
> A clear and concise description of what you expected to happen.

The escaped apostrophe character should have been remained escaped. 

## Describe the bug
> A clear and concise description of what the bug is.

Something this library does transforms escaped apostrophe characters and quotes into their regular counterparts. 

## Details

**Version of Node.js:**
*PLEASE NOTE: Only stable LTS versions (10.x and 12.x) are fully supported but we will do our best with newer versions.*
v23.6.0

**Server Operating System:**
*The server (which might be your dev laptop) on which Apostrophe is running. Linux? MacOS X? Windows? Is Docker involved?*
Windows 

**Additional context:**
> Add any other context about the problem here. If the problem is specific to a browser, OS or mobile device, specify which.

**Screenshots**
*If applicable, add screenshots to help explain your problem.*

<img width="727" height="364" alt="Screenshot of Console" src="https://github.com/user-attachments/assets/97d53f52-ed55-4689-856c-7cdd5b335311" />

PR #718, opened 2025-11-20
chore: add quote and apostrophe test

a2937:
## Summary

This pull request adds two HTML test based on some behavior that seems undesirable. This library automatically turns an escaped apostrophe character into an actual apostrophe character, and same thing with an escaped quote character. 

Related to #717 

Run the test library and notice the error in the console. 

*(Check at least one)*

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [X] Build-related changes
- [ ] Other

- [X] It includes a) the existing issue ID being resolved, b) a convincing reason for adding this feature, or c) a clear description of the bug it resolves
- [X] Related tests have been updated

**Other information:**
I'm not sure the changelog and documentation need updated with this PR alone as none of the internals are changed. As of right now, this realistically shouldn't be merged without a fix.  But as of right now I don't know the code well enough to know how to fix this issue.

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -1848,4 +1848,20 @@ describe('sanitizeHtml', function() {
 
     assert.equal(sanitizedHtmlPreservedAttrsTrue, sanitizedHtmlPreservedAttrsFalse);
   });
+  it('should ignore an apostrophe html entity', () => {
+    const inputHtml = '<code>&apos;</code>';
+    const message = sanitizeHtml(inputHtml, {
+      allowedTags: [ 'b', 'i', 'em', 'strong', 'code', 'wbr' ]
+    });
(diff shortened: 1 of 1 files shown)


boutell:
It is not always necessary to escape ' and " and doing so bloats the markup considerably.

As long as sanitize-html escapes these characters properly when necessary (e.g. inside attributes that, in the output, are quoted with the same character in question), there is no issue.

This library is not intended to preserve the exact style that was used in the original HTML, it only promises to produce valid output which is equivalent (except where it performs sanitization).

boutell:
Thank you for taking the time to make a test to demonstrate your issue. Even if we don't agree on the need for it, we do appreciate the effort.

